PR #21, opened 2015-10-14
More requests added

XyLoNaMiyX:
More requests added:
- GetDialogsRequest
- GetFileRequest
- GetHistoryRequest
- GetNearestDcRequest
- GetUsersRequest

### DIFF
--- a/TLSharp.Core/Requests/GetDialogsRequest.cs
+++ b/TLSharp.Core/Requests/GetDialogsRequest.cs
@@ -1,0 +1,90 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using TLSharp.Core.MTProto;
+
+namespace TLSharp.Core.Requests
+{
+    public class GetDialogsRequest : MTProtoRequest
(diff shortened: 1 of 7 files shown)


sochix:
Please, write here in comment how to use new messages in TelegramClient. Thanks!

XyLoNaMiyX:
```csharp
// Get user
var contactId = await client.ImportContact(phone_number).Result;

// Retrieve messages (get history)
var msgs = await client.GetHistory(contactId, offset, max_id, amount);

// GetDialogs (load available chats and their users). Now you can acceed to client.users and client.chats
await client.LoadChatsAndUsers(offset, max_id, limit);
```

The GetNearestDcRequest has been implemented but not added to TelegramClient.cs. It loads the country, the current DC and the nearest DC.

Same thing happens to GetUsersRequest, it has been implemented but not added to TelegramClient.cs. Given a list of user ids, GetUsersRequest retrieves their users (information such as name, last name, profile photo, etc).

PR #101, opened 2016-02-26
Updated to the latest scheme

XyLoNaMiyX:
.tl files compiler created, everything updated to the latest scheme.
Classes "TBinaryReader" and "TBinaryWriter" created to emulate Telegram
BinaryReader/BinaryWriter


sochix:
Don't understand how actually it compile smthng, describe please.

XyLoNaMiyX:
It takes .tl files (such as [https://core.telegram.org/schema](https://core.telegram.org/schema)) and turns them into CSharp code. Basically, it keeps the structure you've been using until now.

For example:
`inputPeerEmpty#7f3b18ea = InputPeer;`

Turns into:
```csharp
public class InputPeerEmptyType : InputPeer
{
    public override uint ConstructorCode => 0x7f3b18ea;

    public InputPeerEmptyType() { }

    public override void Write(TBinaryWriter writer)
    {
        writer.Write(ConstructorCode);
    }

    public override void Read(TBinaryReader reader)
    {
    }

    public override string ToString()
    {
        return "(InputPeerEmptyType)";
    }
}
```

I made an abstract implementation of a compiler class so any kind of compiler (for example, a compiler for JavaScript could be made) can be made. In this case, I've implemented a CSharp code compiler.

sochix:
What is TBinaryWriter or TBinaryReader in your example?

XyLoNaMiyX:
Instead using a `Serializer` class, I've instead made a class that implements a `BinaryWriter & BinaryReader`, so they act like your old `Serializer` class. They do the same, but makes the code more readable in my opinion.

sochix:
@XyLoNaMiyX pls, try to fix a build

XyLoNaMiyX:
Sorry, what did you mean?

eaba:
@XyLoNaMiyX he meant your pull cannot be merged because it is conflicting with @sochix master branch. Look at what is wrong and correct it

XyLoNaMiyX:
Closing this since the "updated version" of the branch is now pointed out under #197

PR #154, opened 2016-05-16
Added class GetUserFullRequest

chusothe41:
Added class GetUserFullRequest.
Now you can get a UserFull object to work with


sochix:
@chusothe41 pls, write a test on a new functionality and also add this functionality to readme.md.

chusothe41:
@sochix NP, I'll later

PR #171, opened 2016-06-14
Added DeleteContactRequest. Added DeleteContactById and GetUserProfil…

bastianonm:
…ePhoto to TelegramClient, added Helper function that extracts a jpg image from a byte array, with jpeg tags check.


sochix:
Please add a documentation to your func

PR #178, opened 2016-07-19
Fix issues with GetDialogs

woakesd:
Default value of max_id = 0 returns all dialogs
Should return a class with lists of dialog, messages, chats and users.
Add UserForeignConstructor to list of constructors
Add missing detail to DialogConstructor (PeerNotifySettings)
Unpack datastream completely to avoid issue with Ionic exception causing
problems in userForeign parsing
Use more recent code for Dialog construction

### DIFF
--- a/TLSharp.Core/MTProto/TL.cs
+++ b/TLSharp.Core/MTProto/TL.cs
@@ -537,6 +537,9 @@ namespace TLSharp.Core.MTProto
             {0x427425e7, typeof (AudioConstructor)},
             {0x36f8c871, typeof (DocumentEmptyConstructor)},
             {0xf9a39f4f, typeof (DocumentConstructor)},
+            {0xab3a99ac, typeof (DialogConstructor)},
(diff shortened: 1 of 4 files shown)


Jiyuu:
actually i saw the same issues when i tried to use the methods,
should be mentioned that 

 +            {0xab3a99ac, typeof (DialogConstructor)},
 +            {0xd9ccc4ef, typeof (UserRequestConstructor)},
 +            {0x075cf7a8, typeof (UserForeignConstructor)},
these types already exist, simply with different key, not sure why the key is different, probably the keys should just be updated?

woakesd:
I suspect you are right and they should be updated.

The dialog constructor changes at Level 14 when PeerNotifySettings is added. The DialogConstructor read arguably should allow for both because you might be talking to an older server and it's also not apparent if the older dialog can be sent.

https://core.telegram.org/constructor/dialog

sochix:
Thanks! Your code merged.

PR #180, opened 2016-07-21
Add get contacts

woakesd:
I have also fixed a spelling error, two methods called Recieve have been changed to Receive (i before e except after c). Sorry if this seems pedantic!


woakesd:
I've just noticed that the latest layer details published by Telegram are for Layer 23.

I was looking through the code for the CLI to try and understand how to see channels and found that there was an API call that wasn't documented. Then I found a commit in the github log which says it was updated to run Layer 42.

I then had a closer look at messages.getDialogs and that call's id has changed along with the parameters. It doesn't seem worth spending more time on this at the moment!

sochix:
It's ok! Thanks for your work

PR #195, opened 2016-09-06
Changed session store to return null when couldn't load instead of ex…

steavy29:
Returning null instead of exception is better when implementing own ISessionStore. F.e. when we need to provide an empty session for the first run(so just return null instead of throwing artificial exception).
Also this will remove suppressing of all possible exceptions which could be not desired behavior.


sochix:
Thanks!

PR #196, opened 2016-09-06
Added SignUp implementation.

steavy29:
- Added SignUp implementation.
- Replaced exception throwing in Tests when some parameter is not configured in app.config with Debug log. This will make possible launching separate tests without configuring all fields.


sochix:
Thanks!

PR #205, opened 2016-09-13
Add contactName on ImportContact

melvinsalas:
When you import a contact by phone number, by default the name is "My Test Name". I added a optional new parameter (contactName), when it's null set as Default Name


steavy29:
I suggest making this parameter not optional at all. `My Test Name` can be useful only in tests.

armanforghani:
I suggest making this parameter optional but use phone number itself as contact name when it is not provided.

PR #216, opened 2016-09-22
Redesigned TLSharp With Layer 53

aarani:
Hey Everybody ,<br>
 I finally finished this one and i decide to put this in github .<br>
 I Create my own (De)Serializer and Generator To Do this in Best Way .<br>
 Now Layer is Uptodate(Layer 53).
 But For Every Request You Must Wrote your MTProtoRequest Same as what i did for SendCode - SignIn - InitConnection.

Sincerly , ***Afshin Arani***

PR #222, opened 2016-09-26
Update parse userStatus and userStatusConstruction

amastaneh:


sochix:
See https://github.com/sochix/TLSharp/pull/216

PR #223, opened 2016-09-26
Update DocumentConstructor for Layer 23

Tehl:
TL.cs appears to be using the layer 18 schema for Document:

`document#9efc6326 id:long access_hash:long user_id:int date:int file_name:string mime_type:string size:int thumb:PhotoSize dc_id:int = Document;`

however InitConnectionRequest initializes the connection with layer 23. This causes Messages containing Stickers to throw during `TL.Parse()`, since the layer 23 schema for Document is:

`document#f9a39f4f id:long access_hash:long date:int mime_type:string size:int thumb:PhotoSize dc_id:int attributes:Vector<DocumentAttribute> = Document;`

I assume TL.cs is built with an automatic tool, but I couldn't find anything that looked like it matched this output so I've manually added the relevant DocumentAttribute constructors and updated the DocumentConstructor accordingly.

I found that the Document stream for Stickers contains an additional 3 bytes (000) after the `attributes` vector, which I couldn't match to anything in the schema - some kind of padding perhaps? I added a loop to skip over it until the next constructor is found, which seems to do the trick, but I notice this isn't needed anywhere else in the source so I may be missing something here.

With these changes I'm able to use `GetMessagesHistoryForContact` in conversations which contain Stickers; I've not tested with any other api calls as I only have a basic console app at present and I've not implemented any further functionality. Hopefully someone else will find this useful!

### DIFF
--- a/TLSharp.Core/MTProto/TL.cs
+++ b/TLSharp.Core/MTProto/TL.cs
@@ -378,7 +378,7 @@ namespace TLSharp.Core.MTProto
             {0x77bfb61b, typeof (PhotoSizeConstructor)},
             {0xe9a734fa, typeof (PhotoCachedSizeConstructor)},
             {0xc10658a8, typeof (VideoEmptyConstructor)},
-            {0x5a04a49f, typeof (VideoConstructor)},
+            {0x388FA391, typeof (VideoConstructor)},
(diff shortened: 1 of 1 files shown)


sochix:
Thanks to @aarani , now we have the latest TLSchema, and generator for further changes. Thats's why I reject your pull-request

PR #229, opened 2016-10-01
Delete Useless Things from Readme.md

aarani:


sochix:
We need to add description of newly addded func first

aarani:
I Start Documenting in readthedocs.org

PR #231, opened 2016-10-02
Update Readme

aarani:
Remove Unused Date & Add Documention Link


sochix:
Hi! I did not like that you move all docs to your own domain. Let's use github.io for documentation or leave it on Readme.md.

aarani:
It's on github page but problem is that i have personal website on github page so all of my project sites will automatically move them to aarani.ir but we can create organization

PR #255, opened 2016-10-21
build: switch to lower framework version (4.5.2 -> 4.5)

knocte:
This change is good because:
a) It's better to target a lower target framework version if
the project doesn't necessarily depend on the new features
of the newer versions (so, bigger target audience).
b) It lets compile the project with implementations of the
.NET Framework that are not compatible with 4.5.2. For example,
after this change I can successfully build TLSharp with the
Mono v4.2.1 that comes in my Ubuntu Linux 16.04.1.


sochix:
Yep, it's good idea!

PR #256, opened 2016-10-22
Improve exception handling

knocte:
Parsing the message of an exception to decide what to do
next is a bad practice, because it's easy that the message
might be changed by mistake in the future. To enforce the
coupling in a stronger way it's better to use exceptions
of different type depending on the kind of error, so that
we rely on the compiler enforcing the behaviour when doing
changes in this error handling areas in the future.

This also makes the code a bit more simple and readable.


sochix:
Thanks!

sochix:
Added you to https://github.com/sochix/TLSharp/blob/master/README.md#contributors

PR #258, opened 2016-10-23
Tests: put API settings also in app.config

knocte:
For these tests to pass you not only needed to change
the app.config file but the .cs file as well with your
API settings (Hash & ID).

So let's put everything of them in the app.config so
it's easier to discover.


sochix:
Thanks

PR #259, opened 2016-10-23
Tests: avoid NullReferenceException

knocte:
Instead of throwing cryptic NullReferenceException
when this test fails, let's give more information
about what is going on, to the developer.


sochix:
Thanks!

PR #260, opened 2016-10-23
Tests: show a more meaningful exception when running for 1st time

knocte:
Instead of throwing a System.FormatException, capture the URL
thrown by the library itself so that the developer configures
the API_ID and API_HASH and places them in the app.config file.


sochix:
Thanks!

PR #263, opened 2016-10-24
Always include an innerException when re-throwing 

knocte:
Hiding the previous thrown exception is a bad practice.


sochix:
Thanks!

PR #272, opened 2016-10-30
Rework MIGRATE errors/exceptions

knocte:
Several improvements around MIGRATE errors/exceptions


sochix:
Thanks, i like the DataCenterMigrationException abstract class

PR #277, opened 2016-10-30
Friendlier exceptions for Auth testing

knocte:
This change also lets you write the auth code in the
app.config file without the need to recompile/debug.

### DIFF
--- a/TLSharp.Core/Network/MtProtoSender.cs
+++ b/TLSharp.Core/Network/MtProtoSender.cs
@@ -293,6 +293,10 @@ namespace TLSharp.Core.Network
                     var dcIdx = int.Parse(resultString);
                     throw new UserMigrationException(dcIdx);
                 }
+                else if (errorMessage == "PHONE_CODE_INVALID")
+                {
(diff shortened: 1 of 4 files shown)


sochix:
Code sent by Telegram will change frequently. I think it's a bad idea to store it in app.config. Why you want to do it?

knocte:
That is a good question, I figured you would ask it!

Think about the point of view from a person that wants to start contributing to the library: he runs the AuthUser test and it fails because of an PHONE_CODE_INVALID error. He has to go and find the place where the "code" variable is, and you require him to use the debugger to be able to run the test.

It's probably better to make the test fail indicating him already where can he put the code. In fact, I didn't make the code worse, as you can still modify the value of it in the debugger if you want. But this change makes it possible to run the test and make it pass without recompiling or running the debugger (because if you do two SendCode requests very close in time, you will get the same code for both of them, so you can run it once, see it fail, write the code you got in your mobile in your app.config, and run the test again and see it pass).

sochix:
Agreed with you. Thanks for PR and explanation.

knocte:
Thanks for merging! I should have included my explanation in the commit message. Next time, tell me to amend it! :)

PR #284, opened 2016-11-02
SendNumberTest: remove the + from the phone number in app.config

knocte:
NumberToAuthenticate works fine for the AuthTest if we put
a "+" before our international prefix when writing the number,
but because this test compares the numbers we got from the
Contacts List with the one in app.config, it could not find it
due to this prefix. Let's rather be prepared for that.


sochix:
Thanks!

PR #289, opened 2016-11-03
Tests: add NUnit support

knocte:
What's an open source project that cannot be
tested in an open source platform? The best
open source .NET IDE that works in Linux
(MonoDevelop) doesn't have support for
VisualStudio Testing framework, but it has
support for NUnit.

By abstracting a bit the TLSharp.Tests assembly
we can have two different wrappers, one for
NUnit tests and other for VisualStudio tests,
so no one can be left behind.

### DIFF
--- a/TLSharp.Tests.NUnit/TLSharp.Tests.NUnit.csproj
+++ b/TLSharp.Tests.NUnit/TLSharp.Tests.NUnit.csproj
@@ -1,0 +1,54 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project DefaultTargets="Build" ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
+    <ProjectGuid>{E90B705B-19FA-43BA-B952-69957976D12C}</ProjectGuid>
+    <OutputType>Library</OutputType>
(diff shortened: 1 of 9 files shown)


sochix:
I did not like that we have now two test systems. Maybe we can ditch MSTest and use only Nunit? Does NUnit work well on Visual Studio?

knocte:
> Does NUnit work well on Visual Studio?

No idea. I just googled about this and I guess one has to use this: https://visualstudiogallery.msdn.microsoft.com/6ab922d0-21c0-4f06-ab5f-4ecd1fe7175d

knocte:
If you're worried about having to maintain two main suites, I don't mind keeping an eye on TLSharp.Tests.NUnit from time to time to keep it in sync with TLSharp.Tests.VS, in case the latter grows faster than the former.

WRT working with a base and a derived class in each test, it's a bit ugly at first, but very easy to understand: every time you create a new test, you add it in TLSharp.Tests and then later add a wrapper to it in the test suite that you're using.

knocte:
>> Does NUnit work well on Visual Studio?
> No idea. I just googled about this and I guess one has to use this:
> https://visualstudiogallery.msdn.microsoft.com/6ab922d0-21c0-4f06-ab5f-4ecd1fe7175d

Actually, I remember once that actually running NUnit tests in VisualStudio is actually very easy and works very well: when using the Resharper plugin. I used to have it when I worked some years in London for a Microsoft shop (in particular: 7digital). I'm guessing ReSharper is still alive and well, so it's not super-crazy to rely on NUnit. If you want me to create a v3 PR just removing MSTest, let me know.

sochix:
After all I select this version. Thank you, for you work!

PR #294, opened 2016-11-06
Added possibility to do ping request (SendPingAsync method)

timocov:
Hi there,

according to [this](https://github.com/telegramdesktop/tdesktop/blob/master/Telegram/SourceFiles/mtproto/scheme.tl#L109) TL scheme has ping-pong request.

Usage:
```
await client.SendPingAsync();
````

It will send ping request and wait for pong response.


sochix:
Thanks!

PR #295, opened 2016-11-08
Tests: add NUnit support (v2)

knocte:
This is a bit better approach than the previous one done at https://github.com/sochix/TLSharp/pull/289 because the end result is only 2 Tests assemblies, not 3. For more info, see commit message.

### DIFF
--- a/TLSharp.Tests.VS/TLSharpTestsVs.cs
+++ b/TLSharp.Tests.VS/TLSharpTestsVs.cs
@@ -1,0 +1,71 @@
+﻿
+using System.Threading.Tasks;
+
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace TLSharp.Tests
+{
+    [TestClass]
+    public class TLSharpTestsVS : TLSharpTests
+    {
+        [TestInitialize]
+        public void Init()
+        {
+            base.Init(o => Assert.IsNotNull(o), b => Assert.IsTrue(b));
+        }
+        
+        [TestMethod]
+        public override async Task AuthUser()
+        {
+            await base.AuthUser();
+        }
+
+        [TestMethod]
+        public override async Task SendMessageTest()
+        {
+            await base.SendMessageTest();
+        }
+
+        [TestMethod]
+        public override async Task SendMessageToChannelTest()
+        {
+            await base.SendMessageToChannelTest();
+        }
+
+        [TestMethod]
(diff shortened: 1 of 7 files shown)


knocte:
@sochix any feedback?

sochix:
Hi! Can I merge only v2? Will it work w/o v1?

knocte:
Of course! V2 and V1 are independent. You choose which approach you prefer. Once you merge one of them, we can close the other one.

TLSharp.Tests.VS/TLSharpTestsVs.cs
```diff
@@ -0,0 +1,71 @@
+﻿
+using System.Threading.Tasks;
+
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace TLSharp.Tests
+{
+    [TestClass]
+    public class TLSharpTestsVS : TLSharpTests
```
sochix: I want to make TLSharpTests and TLSharpTestsNUnit classes. TlSharpTestsNUnit will be derived from mstest class. So you'll no difference between old tests and new tests. Can we do like this?

knocte: I also thought about doing it this way (inversely) but this is not possible because MSTest-related assemblies are not loaded at all in IDEs which are not Visual Studio. Therefore, you wouldn't be able to compile TLSharpTestsNUnit because it wouldn't compile the referenced project.

aarani: Best Practice is to create SharedLibrary for Tests and call them in NUnit or MSTest

sochix: Correct me if I'm wrong. But it was done by @knocte at v1.

knocte: Exactly, the approach that @aarani describes is the one used in https://github.com/sochix/TLSharp/pull/289 . You guys have the choice of what to merge. I could also create a v3 that just gets rid of the MSTest infrastructure altogether, as @sochix suggested in the other pull request. Please let me know how to proceed. If NUnit support doesn't get merge, IMHO it makes opensource developers like me not really comfortable to work in the library because we depend on a Windows VM running Visual Studio :(

sochix:
V1 selected

PR #296, opened 2016-11-08
Add C# style policy to SLN file

knocte:
This helps MonoDevelop know that the solution will default
to spaces instead of tabs, etc. This doesn't disrupt the
solution in VisualStudio at all.


sochix:
Thanks!

PR #301, opened 2016-11-11
some handy changes

mykolasglinskis:


TLSharp.Core/TelegramClient.cs
```diff
@@ -105,11 +106,63 @@ public bool IsUserAuthorized()
...
+                catch (PhoneMigrationException ex)
+                {
+                    await ReconnectToDcAsync(ex.DC);
+                }
+            }
+
+            return request.Response.phone_registered;
+        }
+
+        public async Task<TLAbsChannelDifference> GetFirst100HundredUpdatesFromTheChannel(TLChannel channel)
```
knocte: This part of the commit is not explained in your commit message, it appears you need to split the commit into two commits.

TLSharp.Tests/TLSharpTests.cs
```diff
@@ -176,6 +177,31 @@ private void GatherTestConfiguration()
...
+        {
+            var client = NewClient();
+
+            await client.ConnectAsync();
+
+            var dialogs = (TLDialogs)await client.GetUserDialogsAsync();
+            var chat = dialogs.chats.lists
+                .Where(c => c.GetType() == typeof(TLChannel))
+                .Cast<TLChannel>()
+                .FirstOrDefault(c => c.title == "Telegram News");
```
sochix: What if user haven't channel with name "Telegram News"?

PR #306, opened 2016-11-16
Add Cloud password Support

aarani:
Hey all.

Now We can use cloud-password enabled accounts on TLSharp. 

Check How to Use in TLSharpTests.cs


TLSharp.Tests/TLSharpTests.cs
```diff
@@ -117,12 +123,18 @@ private void GatherTestConfiguration()
             {
                 user = await client.MakeAuthAsync(NumberToAuthenticate, hash, code);
             }
+            catch (CloudPasswordNeededException ex)
```
aarani: I actually looking forward to find easier way to interact with user to get password but now it's best way

sochix:
Thanks!

PR #307, opened 2016-11-16
Convert TLVector to Enumerable 

aarani:
Why ?  
Using ".lists" for every line of code is very annoying so i just convert TLVector To Enumerable.


knocte:
Why are you doing so many casts? The List is generic so you don't need casts.

aarani:
It's Visual Studio's Generated Code Not Mine

knocte:
Anyway you're mixing things here, you added the CloudPassword support commit.

sochix:
Yep, agreed with @knocte. @aarani can you please factor out cloud passwords from this pull-request?

PR #320, opened 2016-11-29
Introduced a unit test to run phone check for 30 times in order to tr…

mykolasglinskis:
…igger flood exception and "cannot read package size" exception after that

### DIFF
--- a/TLSharp.Tests.NUnit/Test.cs
+++ b/TLSharp.Tests.NUnit/Test.cs
@@ -68,5 +68,11 @@ namespace TLSharp.Tests
         {
             await base.CheckPhones();
         }
+
+        [Test]
+        public override async Task FloodExceptionShouldNotCauseCannotReadPackageLengthError()
+        {
+            await base.FloodExceptionShouldNotCauseCannotReadPackageLengthError();
+        }
     }
 }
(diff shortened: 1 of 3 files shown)


knocte:
What's the purpose of introducing this test?

mykolasglinskis:
the test breaks and something must be fixed in order to make it work. The problem is that from time to time I was getting exception "Cennot read package length" and I thought it is a good ideat to try to reproduce it.

knocte:
Well, it's normal that Telegram bans you if you flood its API. You should just not send so much data at once.

mykolasglinskis:
yes, but after telegrams flood exception (which is not breaking the unit test i wrote) another one happens - "cannot red package length" I think there is a bug.

knocte:
Oh I see.

Can you paste the whole stacktrace of that exception in the commit message please?

knocte:
Also, the test should probably be called `FloodExceptionShouldNotCauseCannotReadPackageLengthError`, to be more clear.

knocte:
Oh, also you need to mark the test as [Ignore("FIXME")] for example, otherwise it will break the test suite when people run it. Last thing, please squash your commits into one, by doing `git rebase -i HEAD~n` (where n is the number of commits you did), and then mark the last ones as "s". Thanks

hosseinGanjyar:
I solved this issue. please send your question in my group:
https://t.me/joinchat/HhLHow3Ed97A0gRZo7Ggog

PR #332, opened 2016-12-07
My implementation of connection through http proxy (with password)

wolvewilson:


knocte:
Hey @wolvewilson thanks for your contribution.

However, I looked at the diff of each commit and it's very difficult to review because, for example, in the first commit, it seems you're using different line-ending format as the original file so we cannot see what you're really changing.

And in the second commit there appears to be many unrelated changes (such as removing exceptions). It seems you need to remove those changes (plus the build of the PR is failing).

wolvewilson:
sorry, it´s my first time :)  i´ll review it today!

wolvewilson:
**Example of use:**

           TelegramClient client;

            bool usaProxy = true;
            int apiId = 0;
            string apiHash = "";
            string apiPhone = "";
            string httpProxyHost = "";
            int httpProxyPort = 0;
            string proxyUsername = "";
            string proxyUserPwd = "";

            if (usaProxy)
                client = new TelegramClient(apiId, apiHash, httpProxyHost, httpProxyPort, proxyUsername, proxyUserPwd);
            else
                client = new TelegramClient(apiId, apiHash);
            await client.ConnectAsync();

knocte:
Can you squash your 4 commits into a single one please? You can do so via `git rebase -i HEAD~4`

Ansel1990B:
There is a bug in ReconnectToDcAsync

This is a bug in transport

Change Code ReconnectToDcAsync:

private async Task ReconnectToDcAsync(int dcId)
        {
            if (dcOptions == null || !dcOptions.Any())
                throw new InvalidOperationException($"Can't reconnect. Establish initial connection first.");

            var dc = dcOptions.First(d => d.id == dcId);

            //_transport = new TcpTransport(dc.ip_address, dc.port);
             _transport = new TcpTransport(_session.ServerAddress, _session.Port,
httpProxyHost, httpProxyPort, proxyUserName, proxyPassword);
            _session.ServerAddress = dc.ip_address;
            _session.Port = dc.port;

            await ConnectAsync(true);
}

To use the proxy, ReconnectToDcAsync should be like the above code

PR #344, opened 2016-12-15
Implementing IDisposable

salarcode:
Disposing the connection is essential.

### DIFF
--- a/TLSharp.Core/TelegramClient.cs
+++ b/TLSharp.Core/TelegramClient.cs
@@ -21,7 +21,7 @@ using TLAuthorization = TeleSharp.TL.Auth.TLAuthorization;
 
 namespace TLSharp.Core
 {
-    public class TelegramClient
+    public class TelegramClient : IDisposable
     {
         private MtProtoSender _sender;
         private AuthKey _key;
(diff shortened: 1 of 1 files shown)


knocte:
Thanks for your contribution! I see that you used tabs in your commits, can you please fix this to use spaces, which is what TLSharp is using at the moment? Thanks

salarcode:
No problem, done.

knocte:
I just noticed you also pushed an extraneous commit not related to the initial PR. Can you move the IsConnected commit to a different PR please?

salarcode:
Actually github sets the whole fork as a pull request, there wasn't anywhere to specify anything specific so this happened!
Only 2 files is changed so that shouldn't be a problem, I think.

knocte:
Not sure you understand. I see the value in having an IDisposable interface if the class has IDisposable fields, so I would merge that change right away. However, I think the introduction of IsConnected requires to have more discussion. That's why it's good you separate them so we can merge one and discuss the other. 

Just remove the commit from your master branch and add it to a new branch to create a new PR for the IsConnected change.

salarcode:
My bad, by PR I though of something else! Done.

knocte:
Ok this PR is almost ready, you just need to squash your commits into one.

You can do this via `EDITOR=nano git rebase -i HEAD~5` and then rename from "pick" to "s" the excess commits (this PR should end up with just one commit after you do this).

salarcode:
Finally managed to squash!

knocte:
@salarcode thanks! but unfortunately the commit message is wrong because you're mentioning the "IsConnected" thing there. You can fix this via `EDITOR=nano git commit --amend`

salarcode:
done

knocte:
Thanks! Merged.

PR #350, opened 2016-12-21
Tests: Use OfType() instead of GetType() == typeof()

salarcode:
There you go!
By the way you take conventions way too seriously that can make some people to give up!
Because I had other changes that I didn't want to commit, I had to remove my local patch and revert them to edit the commit message the way you asked, I doubt that I could do the that next time.

### DIFF
--- a/TLSharp.Tests/TLSharpTests.cs
+++ b/TLSharp.Tests/TLSharpTests.cs
@@ -172,8 +172,7 @@ namespace TLSharp.Tests
             var result = await client.GetContactsAsync();
 
             var user = result.users.lists
-                .Where(x => x.GetType() == typeof(TLUser))
-                .Cast<TLUser>()
+                .OfType<TLUser>()
                 .FirstOrDefault(x => x.phone == normalizedNumber);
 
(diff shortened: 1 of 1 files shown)


knocte:
>There you go!

Thanks

> By the way you take conventions way too seriously that can make some people to give up!

I try to introduce them progressively (i.e.: I wasn't that strict about adding a tag in your first contribution ;) ).

> Because I had other changes that I didn't want to commit, I had to remove my local patch and revert them to edit the commit message the way you asked, I doubt that I could do the that next time.

You wouldn't need to do that, you could have used `git rebase -i HEAD~3` and then use `reword` instead of `pick`. If you ever have troubles with reworking contributions, it means you haven't mastered git enough :)

salarcode:
You're right about git, because I'm not using it as my main source-control and I'm not going to put my time on it anytime soon. Anyway the point is If I were you (*looking at the pull requests history) I wouldn't bother by tags/spaces/line endings. They all are easily maintainable! This whole thing just drives away the motivation to contribute more.

PR #372, opened 2017-01-19
Add offset parameter to GetFile() method

Starli0n:
Hi,

I made this change as it is necessary for downloading larger file.

Indeed after a call to GetFile, TFile.bytes.Length could be lesser than filePartSize which obliges to call it back with an offset.

It has been tested in my code.

Thanks


knocte:
Looks good, thanks!

PR #334, opened 2016-12-08
Connect through http proxy authentication

wolvewilson:
Example of use:

       TelegramClient client;

        bool useProxy = true;
        int apiId = 0;
        string apiHash = "";
        string apiPhone = "";
        string httpProxyHost = "";
        int httpProxyPort = 0;
        string proxyUsername = "";
        string proxyUserPwd = "";

        if (useProxy)
            client = new TelegramClient(apiId, apiHash, httpProxyHost, httpProxyPort, proxyUsername, proxyUserPwd);
        else
            client = new TelegramClient(apiId, apiHash);
        await client.ConnectAsync();

PR #373, opened 2017-01-20
Http proxy authentication

Starli0n:
Hi,

As I needed it, I implemented it based on pull #334 and pull #332.
Thank @wolvewilson
But this time with a clean diff :-)

To his discharge, there are some inconsistent line ending in some files.
(ie: a mix between Window CRLF and Linux LF end of files and also a mix between tab and spaces for indentation)

It works perfectly with my project.

Thanks

### DIFF
--- a/TLSharp.Core/Network/TcpTransport.cs
+++ b/TLSharp.Core/Network/TcpTransport.cs
@@ -1,22 +1,28 @@
 ﻿using System;
-using System.Linq;
 using System.Net;
 using System.Net.Sockets;
 using System.Threading.Tasks;
 
 namespace TLSharp.Core.Network
 {
+    public delegate TcpClient TcpClientConnectionHandler(string address, int port);
+
     public class TcpTransport : IDisposable
     {
         private readonly TcpClient _tcpClient;
         private int sendCounter = 0;
 
-        public TcpTransport(string address, int port)
+        public TcpTransport(string address, int port, TcpClientConnectionHandler handler = null)
         {
-            _tcpClient = new TcpClient();
+            if (handler == null)
+            {
+                _tcpClient = new TcpClient();
 
-            var ipAddress = IPAddress.Parse(address);
-            _tcpClient.Connect(ipAddress, port);
+                var ipAddress = IPAddress.Parse(address);
+                _tcpClient.Connect(ipAddress, port);
+            }
+            else
+                _tcpClient = handler(address, port);
         }
 
         public async Task Send(byte[] packet)
--- a/TLSharp.Core/TelegramClient.cs
+++ b/TLSharp.Core/TelegramClient.cs
@@ -28,8 +28,10 @@ namespace TLSharp.Core
         private int _apiId = 0;
         private Session _session;
         private List<TLDcOption> dcOptions;
+        private TcpClientConnectionHandler _handler;
 
-        public TelegramClient(int apiId, string apiHash, ISessionStore store = null, string sessionUserId = "session")
+        public TelegramClient(int apiId, string apiHash,
+            ISessionStore store = null, string sessionUserId = "session", TcpClientConnectionHandler handler = null)
         {
             if (apiId == default(int))
                 throw new MissingApiConfigurationException("API_ID");
@@ -42,9 +44,10 @@ namespace TLSharp.Core
             TLContext.Init();
             _apiHash = apiHash;
             _apiId = apiId;
+            _handler = handler;
 
             _session = Session.TryLoadOrCreateNew(store, sessionUserId);
-            _transport = new TcpTransport(_session.ServerAddress, _session.Port);
+            _transport = new TcpTransport(_session.ServerAddress, _session.Port, _handler);
         }
 
         public async Task<bool> ConnectAsync(bool reconnect = false)
@@ -85,7 +88,7 @@ namespace TLSharp.Core
 
             var dc = dcOptions.First(d => d.id == dcId);
 
-            _transport = new TcpTransport(dc.ip_address, dc.port);
+            _transport = new TcpTransport(dc.ip_address, dc.port, _handler);
             _session.ServerAddress = dc.ip_address;
             _session.Port = dc.port;
 


TLSharp.Core/Network/TcpTransport.cs
```diff
@@ -84,6 +91,53 @@ public void Dispose()
         {
             if (_tcpClient.Connected)
                 _tcpClient.Close();
+        }
+
+        static TcpClient connectViaHTTPProxy(string targetHost, int targetPort,
```
knocte: Please use `ConnectViaHttpProxy` as the name of the method, to respect .NET API design guidelines.

TLSharp.Core/Network/TcpTransport.cs
```diff
@@ -11,12 +12,18 @@ public class TcpTransport : IDisposable
         private readonly TcpClient _tcpClient;
         private int sendCounter = 0;
 
-        public TcpTransport(string address, int port)
+        public TcpTransport(string address, int port,
+            string httpProxyHost = "", int httpProxyPort = 0, string proxyUserName = "", string proxyPassword = "")
```
knocte: Instead of passing 4 parameters, can you wrap these elements into a class (e.g. ProxyInfo?) that contains two properties: Uri (this way it would include host and port) and NetworkCredentials (which includes user and password).

knocte: Or rather, just use the already existing WebProxy class.

TLSharp.Core/TelegramClient.cs
```diff
@@ -28,8 +28,13 @@ public class TelegramClient : IDisposable
...
         private Session _session;
         private List<TLDcOption> dcOptions;
+        private string _httpProxyHost;
+        private int _httpProxyPort;
+        private string _proxyUserName;
+        private string _proxyPassword;
 
-        public TelegramClient(int apiId, string apiHash, ISessionStore store = null, string sessionUserId = "session")
+        public TelegramClient(int apiId, string apiHash, ISessionStore store = null, string sessionUserId = "session",
+            string httpProxyHost = "", int httpProxyPort = 0, string proxyUserName = "", string proxyPassword = "")
```
knocte: Same here, use WebProxy class instead please.

Starli0n:
Ok I integrated all your remarks and it is working!

Thank you, I learnt 👍

TLSharp.Core/TelegramClient.cs
```diff
@@ -42,9 +45,29 @@ public TelegramClient(int apiId, string apiHash, ISessionStore store = null, str
...
             _apiHash = apiHash;
             _apiId = apiId;
+            _webProxy = webProxy;
 
             _session = Session.TryLoadOrCreateNew(store, sessionUserId);
-            _transport = new TcpTransport(_session.ServerAddress, _session.Port);
+            _transport = new TcpTransport(_session.ServerAddress, _session.Port, _webProxy);
+        }
+
+        static public WebProxy CreateWebProxy(string httpProxyHost, int httpProxyPort = 80,
```
knocte: Why do you add this method if you're not using it anywhere? Seems like you can delete it.

TLSharp.Core/Network/TcpTransport.cs
```diff
@@ -1,7 +1,8 @@
 ﻿using System;
-using System.Linq;
+using System.Diagnostics;
 using System.Net;
 using System.Net.Sockets;
+using System.Reflection;
```
knocte: Why are you using Reflection?

TLSharp.Core/Network/TcpTransport.cs
```diff
@@ -84,6 +90,42 @@ public void Dispose()
         {
             if (_tcpClient.Connected)
                 _tcpClient.Close();
+        }
+
+        static TcpClient ConnectViaHTTPProxy(string targetHost, int targetPort, WebProxy webProxy)
```
knocte: It's better to use `Http`, not `HTTP`, to be consistent with the class names in the .NET framework, e.g.: https://msdn.microsoft.com/en-us/library/system.web.httprequest(v=vs.110).aspx

TLSharp.Core/Network/TcpTransport.cs
knocte: Ok, seems like the original contributor used this: http://stackoverflow.com/questions/17751835/connecting-to-a-tcp-server-via-tcpclient-through-proxy

However, using Reflection this way is very likely to cause it to not work with Mono (e.g. in Linux/Mac), so I would prefer if you implement it this way: http://stackoverflow.com/questions/18289418/c-sharp-tcpclient-connect-via-a-proxy

TLSharp.Core/TelegramClient.cs
Starli0n: It was to provide a sample of code to initialize a WebProxy from the parameters but I removed it

TLSharp.Core/Network/TcpTransport.cs
Starli0n: I renamed the method `ConnectViaHttpPProxy`

TLSharp.Core/Network/TcpTransport.cs
Starli0n: The sample of code does not work for me :-/
I do not know what to do with it...

Starli0n:
As I did not succeed in implementing the method you gave me for the http proxy connection, I offer you to resolve that issue by including a delegate to the connection.

Therefore, the client has the choice to provide the library with its own connection, in my case made with a web proxy.

I hope it would be fine with you.

TLSharp.Core/Network/TcpTransport.cs
```diff
@@ -8,15 +7,22 @@ namespace TLSharp.Core.Network
 {
     public class TcpTransport : IDisposable
     {
+        public delegate TcpClient TcpClientHandler(string address, int port);
```
knocte: Given that this handler would replace entirely the otherwise `tcpClient.Connect()` call, maybe it's better to call it TcpClientConnectionHandler.

TLSharp.Core/TelegramClient.cs
```diff
@@ -15,6 +15,7 @@
 using TLSharp.Core.MTProto.Crypto;
 using TLSharp.Core.Network;
 using TLSharp.Core.Utils;
+using static TLSharp.Core.Network.TcpTransport;
```
knocte: I would prefer you place the handler outside the class, to avoid this weird `using` line :)

TLSharp.Core/TelegramClient.cs
```diff
@@ -28,8 +29,10 @@ public class TelegramClient : IDisposable
         private int _apiId = 0;
         private Session _session;
         private List<TLDcOption> dcOptions;
+        private TcpClientHandler _handler;
 
-        public TelegramClient(int apiId, string apiHash, ISessionStore store = null, string sessionUserId = "session")
+        public TelegramClient(int apiId, string apiHash, ISessionStore store = null, string sessionUserId = "session",
+            TcpClientHandler handler = null)
```
knocte: If you break this line in two, it feels better to place all mandatory parameters in one line, and all optional parameters in the next line, you mind doing that?

knocte:
And last but not least: if you're using this approach, it would be better to change the commit message from `Http proxy authentication` to `Custom tcp hanlder to allow HTTP proxy auth`.

Starli0n:
I performed all your requests.

Hope it is good now

TLSharp.Core/TelegramClient.cs
```diff
@@ -28,8 +28,10 @@ public class TelegramClient : IDisposable
         private int _apiId = 0;
         private Session _session;
         private List<TLDcOption> dcOptions;
+        private TcpClientConnectionHandler _handler;
 
-        public TelegramClient(int apiId, string apiHash, ISessionStore store = null, string sessionUserId = "session")
+        public TelegramClient(int apiId, string apiHash,
+            ISessionStore store = null, string sessionUserId = "session", TcpClientConnectionHandler handler = null)
```
knocte: You forgot what I said about params here ;)

Starli0n: For me, all mandatory parameters are on the 1st

- apiId, apiHash

Optional parameters on a 2nd line

- store, sessionUserId, handler

I might not get what you asked...

knocte: Oh oops, overlooked, my bad.

TLSharp.Core/TelegramClient.cs
```diff
@@ -28,8 +28,10 @@ public class TelegramClient : IDisposable
         private int _apiId = 0;
         private Session _session;
         private List<TLDcOption> dcOptions;
+        private TcpClientConnectionHandler _handler;
```
knocte: Only one last thing: is this field really needed? it seems you only use it to pass it to the TcpTransport constructor.

knocte: Nevermind, I hadn't seen the other usage. Merging...

Starli0n: Yes this field is required for `ReconnectToDcAsync()` method

Starli0n:
Thank you for the merge 👍

Harpagon:
@Starli0n how to use this implementation with HTTP proxy auth?
Please, give an example of use.

PR #399, opened 2017-02-13
Some generic exceptions refactored into custom and specific exceptions

mykolasglinskis:


knocte:
The patch is good in general, but I don't see why these exceptions should be public, can you make them internal?

PR #410, opened 2017-02-24
Some generic exceptions refactored into custom and specific exceptions

mykolasglinskis:


knocte:
Thanks for your contribution! You don't need the `$` character of strings because you're not using string interpolation in them. If you remove that I'll merge this patch.

knocte:
(PS: You don't need to close this PR and open a new one to be able to make changes.)

mykolasglinskis:
removed interpolation

knocte:
Your last commit is introducing unnecessary diff noise, did you change the line endings or something?

knocte:
hello?

knocte:
@mykolasglinskis ping?

MasihAkbari:
@knocte so the original poster wont answer anymore so i propose to close this PR and i Open another with the same changes so PR can merge into project. what do you think

knocte:
@MasihAkbari yeah sure, we can close this one when you have opened a new one.

PR #412, opened 2017-03-02
Implementing IsConnected

salarcode:


knocte:
Why does this PR contain so many commits? I guess you merged instead of rebasing.

PR #435, opened 2017-04-03
PCL Support

felix-d-git:
It would be very nice if PCL would be supported.
So this can be used for XAMARIN and others.


knocte:
Old PR?

PR #438, opened 2017-04-04
Improvement on TL converter capability

hadi77ir:
As you know, Telegram is not updating its official version of TL schema (nor TL neither JSON). We only have access to latest schema through "Telegram Desktop" repository and TJ Horner website. The bad news is: it is in TL language format and not JSON.

Be happy! From now, you can use TeleSharp.Generator for these conversions:
1. TL to JSON
2. TL to C#
3. JSON to C#

Also, it now has support for command-line options through NDesk.Options.

For the record, latest TL schema is here: http://tjhorner.com/tl-schema/

### DIFF
--- a/TeleSharp.Generator/Models.cs
+++ b/TeleSharp.Generator/Models.cs
@@ -8,29 +8,98 @@ namespace TeleSharp.Generator
 {
     class Method
     {
-        public int id { get; set; }
-        public string method { get; set; }
+        [Newtonsoft.Json.JsonProperty("id")]
+        public int Id { get; set; }
+
+        [Newtonsoft.Json.JsonProperty("method")]
+        public string MethodName { get; set; }
+
         [Newtonsoft.Json.JsonProperty("params")]
         public List<Param> Params { get; set; }
-        public string type { get; set; }
 
+        [Newtonsoft.Json.JsonProperty("type")]
+        public string ResponseType { get; set; }
+
+        public override bool Equals(Object obj)
+        {
+            return obj is Method && this == (Method)obj;
+        }
+        public override int GetHashCode()
+        {
+            return Id.GetHashCode() ^ MethodName.GetHashCode() ^ Params.GetHashCode() ^ ResponseType.GetHashCode();
+        }
+        public static bool operator ==(Method x, Method y)
+        {
+            return Enumerable.SequenceEqual(x.Params, y.Params) && x.MethodName == y.MethodName && x.ResponseType == y.ResponseType && x.Id == y.Id;
+        }
+        public static bool operator !=(Method x, Method y)
+        {
+            return !(x == y);
+        }
     }
     class Param
     {
-        public string name { get; set; }
-        public string type { get; set; }
+        [Newtonsoft.Json.JsonProperty("name")]
+        public string Name { get; set; }
+
+        [Newtonsoft.Json.JsonProperty("type")]
+        public string Type { get; set; }
     }
     class Constructor
     {
-        public int id { get; set; }
-        public string predicate { get; set; }
+        [Newtonsoft.Json.JsonProperty("id")]
+        public int Id { get; set; }
+
+        [Newtonsoft.Json.JsonProperty("predicate")]
+        public string ConstructorName { get; set; }
+
         [Newtonsoft.Json.JsonProperty("params")]
         public List<Param> Params { get; set; }
-        public string type { get; set; }
+
+        [Newtonsoft.Json.JsonProperty("type")]
+        public string BaseType { get; set; }
+
+        public override bool Equals(Object obj)
+        {
+            return obj is Constructor && this == (Constructor)obj;
+        }
+        public override int GetHashCode()
+        {
+            return Id.GetHashCode() ^ ConstructorName.GetHashCode() ^ Params.GetHashCode() ^ BaseType.GetHashCode();
+        }
+        public static bool operator ==(Constructor x, Constructor y)
+        {
+            return Enumerable.SequenceEqual(x.Params, y.Params) && x.ConstructorName == y.ConstructorName && x.BaseType == y.BaseType && x.Id == y.Id;
+        }
+        public static bool operator !=(Constructor x, Constructor y)
+        {
+            return !(x == y);
+        }
     }
     class Schema
     {
-        public List<Constructor> constructors { get; set; }
-        public List<Method> methods { get; set; }
+        [Newtonsoft.Json.JsonProperty("constructors")]
+        public List<Constructor> Constructors { get; set; }
+
+        [Newtonsoft.Json.JsonProperty("methods")]
+        public List<Method> Methods { get; set; }
+
+        public override bool Equals(Object obj)
+        {
+            return obj is Schema && this == (Schema)obj;
+        }
+        public override int GetHashCode()
+        {
+            return Constructors.GetHashCode() ^ Methods.GetHashCode();
+        }
+        public static bool operator ==(Schema x, Schema y)
+        {
+            return Enumerable.SequenceEqual(x.Methods.OrderBy(t=>t.MethodName),y.Methods.OrderBy(t => t.MethodName)) &&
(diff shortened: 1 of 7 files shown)


TeleSharp.Generator/Program.cs
```diff
@@ -16,39 +17,195 @@ class Program
...
+                            TargetNamespace = m.Groups[0].Value;
+                        }
+                    }
+                    else
+                    {
+                        invalidTargetNamespace = true;
+                    }
+                })
+                .Add("output-json", a => outputJson = a != null)
+                .Add("template-abstract=", a => AbsStyle = (a != null) ? File.ReadAllText(a) : AbsStyle)
```
hadi77ir: A good thing I added was the custom template. I think it will be usable to someone who wants to develop his/her own client library.

TeleSharp.Generator/Program.cs
```diff
@@ -200,11 +376,36 @@ static void Main(string[] args)
...
+                output += $"        public {property.type} {property.name} {{ get; set; }}" + Environment.NewLine;
+            }
+            return output;
         }
+
         public static string FormatName(string input)
         {
             if (String.IsNullOrEmpty(input))
-                throw new ArgumentException("ARGH!");
+                throw new ArgumentException("ARGH! Class Name was empty.");
```
hadi77ir: I thought the exception message didn't explain anything, so I decided to add some words in there to clarify the throwing reason.

TeleSharp.Generator/Program.cs
```diff
@@ -298,15 +499,16 @@ public static string GetTypeName(string type)
                     return "TLObject";
                 case "x":
                     return "TLObject";
+                case "vector t":
+                    return "List<T>";
             }
 
-            if (type.StartsWith("Vector"))
+            if (type.StartsWith("Vector<"))
```
hadi77ir: The previous code didn't work with "vector t" type, I made it like this to make it work with "vector t" type.

TeleSharp.Generator/TL2JSON.cs
```diff
@@ -0,0 +1,110 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Text.RegularExpressions;
+using System.Threading.Tasks;
+
+namespace TeleSharp.Generator
+{
+    static class TL2JSON
```
hadi77ir: This class is the heart of TL language translation. I wrote it using Regex, although I had no experience in Regex.

hadi77ir:
One of the things I have done is moving fields to base classes. This way there is little to no need for casting types or using reflection to get and set properties. (As I had to do in my project, but I have no longer need to cast objects)
This is the thing they have done in official Android app.

TeleSharp.Generator/Program.cs
```diff
@@ -429,4 +631,18 @@ public static FileStream MakeFile(string path)
         }
     }
 
+    struct Property
+    {
+        public string type;
```
knocte: Can you use PascalCase (instead of lowercase/camelCase) for public members, to follow .NET coding guidelines?

Also better add a constructor that receives both type and name values as parameters of it.

TeleSharp.Generator/TL2JSON.cs
```diff
@@ -0,0 +1,110 @@
...
+                me =>
+                {
+                    if (me.Value.StartsWith("/*") || me.Value.StartsWith("//"))
+                        return me.Value.StartsWith("//") ? Environment.NewLine : "";
+                    // Keep the literal strings
+                    return me.Value;
+                },
+                RegexOptions.Singleline);
+        }
+        public static string RemoveEmptyLines(string input) => Regex.Replace(input, @"^\s+$[\r\n]*", "", RegexOptions.Multiline);
```
knocte: Can you split the line in two? It's too long.

TeleSharp.Generator/TeleSharp.Generator.csproj
```diff
@@ -69,6 +86,18 @@
...
+      <Visible>False</Visible>
+      <ProductName>Microsoft .NET Framework 4.5 %28x86 and x64%29</ProductName>
+      <Install>true</Install>
+    </BootstrapperPackage>
+    <BootstrapperPackage Include="Microsoft.Net.Framework.3.5.SP1">
+      <Visible>False</Visible>
+      <ProductName>.NET Framework 3.5 SP1</ProductName>
+      <Install>false</Install>
+    </BootstrapperPackage>
+  </ItemGroup>
```
knocte: Why do you need these changes to the project file?

TeleSharp.Generator/TeleSharp.Generator.csproj
```diff
@@ -12,6 +12,21 @@
...
+    <UpdateInterval>7</UpdateInterval>
+    <UpdateIntervalUnits>Days</UpdateIntervalUnits>
+    <UpdatePeriodically>false</UpdatePeriodically>
+    <UpdateRequired>false</UpdateRequired>
+    <MapFileExtensions>true</MapFileExtensions>
+    <ApplicationRevision>0</ApplicationRevision>
+    <ApplicationVersion>1.0.0.%2a</ApplicationVersion>
+    <IsWebBootstrapper>false</IsWebBootstrapper>
+    <UseApplicationTrust>false</UseApplicationTrust>
+    <BootstrapperEnabled>true</BootstrapperEnabled>
```
knocte: Why do you need these changes to the project file?

knocte:
Hey Mohammad! This contribution is great. I wonder if you can write some tests so that we can see the exact test cases that your Regexes are solving, and so that it doesn't get broken accidentally in the future?

hadi77ir:
I tested its functionality with the latest schema and it works! Its output was good, except for TLVector which has to be written and implemented by hand. (may I have to exclude TLVector from Code generation?)

TeleSharp.Generator/TeleSharp.Generator.csproj
hadi77ir: It seems I have accidentally clicked "Publish" and these changes happened to project file.

hadi77ir:
A good thing about this commit is that it helps other projects that need the schema in JSON format but can't find it anywhere, with this commit they can convert TL to JSON.

TeleSharp.Generator/Program.cs
knocte: Do you mind proposing this single change as a separate PR please?

TeleSharp.Generator/Program.cs
knocte: Do you mind proposing this single change as a separate PR please?

knocte:
> Also, I have to say I know nothing about Unit Testing

You didn't know anything about Regex either ;-) There are some unit tests already in the repo, you can play with them :)

hadi77ir:
The only problem now is: It can't translate [Telegram Desktop Scheme](https://github.com/telegramdesktop/tdesktop/blob/master/Telegram/Resources/scheme.tl) to JSON. I'm currently working on it, but it now works very well with [TJ Horner's version of TL Scheme](http://tjhorner.com/tl-schema/) (which is good enough for now).

hadi77ir:
Now It works with [Scheme from Telegram Desktop Repository](https://github.com/telegramdesktop/tdesktop/tree/master/Telegram/Resources/scheme.tl) which will help our project to move with latest features of Telegram (such as payment, calls, help etc.)

TeleSharp.Generator/Models.cs
```diff
@@ -8,29 +8,98 @@ namespace TeleSharp.Generator
 {
     class Method
     {
-        public int id { get; set; }
-        public string method { get; set; }
+        [Newtonsoft.Json.JsonProperty("id")]
```
hadi77ir: As I found Newtonsoft.Json attributes, I thought their use will make code more readable. 😀

hadi77ir:
I'm thinking after testing the output I have to remove/close this PR for now, 'til issues get fixed (I have to find a way to find conflicting params and add them to inherited classes instead of abstract class)

aarani:
It’s not needed at all because latest Json Schema are always accessable from webogram github and my gist too.

PR #446, opened 2017-04-13
Add length validation foe caption

omarmaher100:
Hi Geeks 
I want to add validation on caption length because the maximum length of it is 200 character


knocte:
Hi Omar, thanks for your contribution.

I say many whitespace changes in the diff, can you remove them please? With them, it's not clear what you're changing.

PR #448, opened 2017-04-13
Add caption length validation

omarmaher100:
Hi Geeks
I want to add validation on caption length because the maximum length of it on telegram server is 200 character

### DIFF
--- a/TeleSharp.TL/TL/TLInputMediaUploadedDocument.cs
+++ b/TeleSharp.TL/TL/TLInputMediaUploadedDocument.cs
@@ -22,7 +22,7 @@ namespace TeleSharp.TL
      public TLAbsInputFile file {get;set;}
      public string mime_type {get;set;}
      public TLVector<TLAbsDocumentAttribute> attributes {get;set;}
-        private string caption;
+     private string caption;
      public string Caption
         {
             get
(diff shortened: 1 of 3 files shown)


TeleSharp.TL/TL/TLInputMediaUploadedDocument.cs
```diff
@@ -22,7 +22,8 @@ public override int Constructor
      public TLAbsInputFile file {get;set;}
      public string mime_type {get;set;}
      public TLVector<TLAbsDocumentAttribute> attributes {get;set;}
-     public string caption {get;set;}
+        private string caption;
```
knocte: Wrong indentation, please use same indentation as before.

TeleSharp.TL/TL/TLInputMediaUploadedDocument.cs
```diff
@@ -22,7 +22,8 @@ public override int Constructor
...
+     public string Caption { get { return caption; } set { caption = value.Length > 200 ? value.Remove(199, value.Length - 199) : value; } }
```
knocte: This line being so long is not readable, please split in many lines.

knocte:
Where is the maximum length documented?

knocte:
Why do you close the PR? You can change your commits without the need to create a new PR.

omarmaher100:
sorry i don't know how :)

PR #449, opened 2017-04-13
caption length validation

omarmaher100:
Hi Geeks
I want to add validation on caption length because the maximum length of it is 200 character
and you can find that here https://core.telegram.org/bots/api just find caption word

### DIFF
--- a/TLSharp.Core/TelegramClient.cs
+++ b/TLSharp.Core/TelegramClient.cs
@@ -269,7 +269,7 @@ namespace TLSharp.Core
                 random_id = Helpers.GenerateRandomLong(),
                 background = false,
                 clear_draft = false,
-                media = new TLInputMediaUploadedPhoto() { file = file, caption = caption },
+                media = new TLInputMediaUploadedPhoto() { file = file, Caption = caption },
                 peer = peer
             });
         }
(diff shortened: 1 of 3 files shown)


knocte:
Why do you think the correct fix here is to truncate the caption?

You could catch the MEDIA_CAPTION_TOO_LONG error, wrap it into a new MediaCaptionTooLongException, put an understandable message for the developer, and be done.

This way, we would push the problem to the consumer of TLSharp. Ideally, the error reporting on input validation should happen in the UI. If you truncate the message, the developer will never know about the problem. Let the developer decide if he wants to truncate the message or tell the user of his app to make a shorter message, don't you think?

omarmaher100:
Ok but why i will wait to get MEDIA_CAPTION_TOO_LONG from telegram although i know the maximum length and can make it in short way and make something like this 
is it ok ?
```
            get
            {
                return caption;
            }
            set
            {
                if (value.Length > 200)
                {
                    throw new System.ArgumentException("Too long caption please make it less than 200 character");
                }
                else
                {
                    caption = value;
                }
            }
```

knocte:
> but why i will wait to get MEDIA_CAPTION_TOO_LONG from telegram although i know the maximum length and can make it in short

If you know the maximum, and you're ok with making it shorter in your application instead of reporting the problem to the user, go ahead! Do the truncate operation in your app. No need to put this logic in TLSharp.

omarmaher100:
No i mean i will not cut any thing just i will tell user by the my exception not the telegram exception

knocte:
Your code should catch the MEDIA_CAPTION_TOO_LONG error, and only then throw the appropiate exception. If you do this, then TLSharp will not need to change if in the future Telegram decides to remove this limitation.

PR #460, opened 2017-04-24
Implementing IsConnected

salarcode:


TLSharp.Core/TelegramClient.cs
```diff
@@ -377,8 +377,15 @@ private void OnUserAuthenticated(TLUser TLUser)
             _session.SessionExpires = int.MaxValue;
 
             _session.Save();
-        }
-
+        }
+
+        public bool IsConnected()
```
knocte: Why do you make this a method and the other a property? I was expecting some consistency here :)

salarcode:
Don't know why implementing connection KeepAlive commit got it in the way. Any was KeepAlive implementation is on the way.

TLSharp.Core/Network/TcpTransport.cs
```diff
@@ -12,7 +12,7 @@ public class TcpTransport : IDisposable
         private readonly TcpClient _tcpClient;
         private int sendCounter = 0;
 
-        public TcpTransport(string address, int port, TcpClientConnectionHandler handler = null)
+        public TcpTransport(string address, int port, bool keepAlive = false, TcpClientConnectionHandler handler = null)
```
knocte: Can you separate this commit into its own PR please?

salarcode: Github did it automatically. How to fix it?

TLSharp.Core/TelegramClient.cs
salarcode: Because it is a resource intensive inquiry that needed to be a method.

TLSharp.Core/Network/TcpTransport.cs
knocte: github cannot push code from you by mistake :) you pushed it to the wrong branch ;)

TLSharp.Core/TelegramClient.cs
knocte: Resource demanding? It's only checking for null and then later returning a property.

salarcode: Yes, resource demanding underneath TcpClient

salarcode:
I have reverted KeepAlive commit for you to have a clean merge.

TLSharp.Core/TelegramClient.cs
salarcode: Also there is more change coming to this in next commit.

knocte: What do you mean? The method as it is now is not resource demanding so please make it a property. If in a second PR it becomes resource demanding, then convert it to a method in the next PR ;)

salarcode: Ok then. dealing with your policy of clean commit takes time. Maybe another time.

knocte: @salarcode I've fixed it myself and merged manually, feel free to propose the KeepAlive PR now ;)

Issue #10, opened 2015-10-01
Connection KeepAlive


delphiassistant:
Hi Again.
I'm developing a windows application which serves a WCF Service, which other apps could invoke it for telegram messaging (instead of sms notifications to my users).

After a while since application is idle, it seems that tcp connection get closed and I get errors when calling sendmessage method.
Do have any advice for this problem?

sochix:
See TcpTransport class, don't know a quck solution for this problem. If you found one, please make a pull-request

eaba:
how about this http://www.codeproject.com/Articles/117557/Set-Keep-Alive-Values

hosseinGanjyar:
I have got same issue. I using WCF and I want to alive connections. In below I show my code:
```
var session = new FileSessionStore();
TelegramClient client = new TelegramClient(apiId, apiHash, session, "session_N");
await client.ConnectAsync();
```
How do I using  **client** for reusable it in many times?
@eaba , @delphiassistant  pleasehelp me if you know any thing.
@eaba I have read above article of you, but I haven't understood. Please show a sample for my code.
**Thanks**

PR #468, opened 2017-05-03
Implementing keep-alive

salarcode:


TLSharp.Core/Network/TcpTransport.cs
```diff
@@ -84,7 +91,58 @@ public TcpTransport(string address, int port, TcpClientConnectionHandler handler
...
+                    SIO_KEEPALIVE_VALS[i * bytesperlong + 1] = (byte)(input[i] >> ((bytesperlong - 3) * bitsperbyte) & 0xff);
+                    SIO_KEEPALIVE_VALS[i * bytesperlong + 0] = (byte)(input[i] >> ((bytesperlong - 4) * bitsperbyte) & 0xff);
+                }
+                // create bytestruct for result (bytes pending on server socket)
+                byte[] result = BitConverter.GetBytes(0);
+
+                // write SIO_VALS to Socket IOControl
+                sock.IOControl(IOControlCode.KeepAliveValues, SIO_KEEPALIVE_VALS, result);
+            }
+            catch (Exception)
```
knocte: what exception type are you actually expecting here?

salarcode: All exceptions of IOControl and possibility of an overflow which is not essentially required. It is a fail-safe catch.

knocte: what exceptions are documented to be thrown by IOControl?

salarcode: https://msdn.microsoft.com/en-us/library/ms145152(v=vs.110).aspx

knocte: You should make the try-catch smaller then? Only focusing on the IOControl call.

salarcode: Current implementation is good enough.

knocte: If somebody in the future introduces a change around the array filling code, and his code is wrong for some reason (e.g. it throws IndexOutOfRangeException), then due to your bigger try-catch, we won't notice the problem.

salarcode: That's another reason. Fair enough. Change it as you wish.

knocte: Can you provide a link where it explains what you're doing with all those byte flags? I'm guessing you extraced the code from somewhere.

TLSharp.Core/TelegramClient.cs
```diff
@@ -110,15 +112,15 @@ public bool IsUserAuthorized()
 
             var authCheckPhoneRequest = new TLRequestCheckPhone() { phone_number = phoneNumber };
             var completed = false;
-            while(!completed)
+            while (!completed)
```
knocte: this change is not needed, please revert

TLSharp.Core/TelegramClient.cs
```diff
@@ -110,15 +112,15 @@ public bool IsUserAuthorized()
...
             {
                 try
                 {
                     await _sender.Send(authCheckPhoneRequest);
                     await _sender.Receive(authCheckPhoneRequest);
                     completed = true;
                 }
-                catch(PhoneMigrationException e)
+                catch (PhoneMigrationException e)
```
knocte: this change is not needed, please revert

TLSharp.Core/TelegramClient.cs
salarcode: Not my fault, Autoformat does it with default VS settings :)

TLSharp.Core/Network/TcpTransport.cs
salarcode: That is an old code which I have used and altered a little. This is the closest one that I found if that helps.
http://codegists.com/snippet/c/net-socket-keep-alive-extension
deeper info: 
https://blogs.technet.microsoft.com/nettracer/2010/06/03/things-that-you-may-want-to-know-about-tcp-keepalives/

salarcode:
About format issues, it seems you are using Unix EOL which is bizarre for a C# window programmer ;)
You can merge the commit the reformat it as you wish.

TLSharp.Core/TelegramClient.cs
knocte: Even if not your fault, if you want to make whitespace changes you have to propose them in a different commit or PR ;)

TLSharp.Core/Network/TcpTransport.cs
knocte: Great, can you add a comment with those URLs please? Also, note that the code from the last URL uses a catch with SocketException, not Exception, so I think it's better in the way it's posted there.

salarcode: Those are not the original source, hence the code search. Also the exception doesn't cover other possible exceptions.

TLSharp.Core/TelegramClient.cs
salarcode: I thought you were sensitive about clean commits, several commits and etc.

knocte: That's exactly the reason why I'm asking you to clean the commit from noise.

salarcode: Come on, all these discussions for a space and EOL? I'm sure doing it would be much easier for you than me.

knocte: If you don't go through the pain of fixing your pull requests and always rely on me, you will never propose clean PRs ;)

knocte: BTW this is a good read: http://tirania.org/blog/archive/2010/Dec-31.html

salarcode: Since we got here, a few things needs be cleared up:
- First, I'm contributing because I have made some changes to the library that I have used in my project and I thought everyone could benefit from it. This doesn't mean that I'm dedicated my time to work on that library anymore. Also it means that my focus is on the features rather than consistency.
- Second, reformatting/resetting/reverting/etc takes more time for me than you as the owner, because I don't have the environment and the tools/settings as you do. I as a contributor should try to match the original guidelines for sure, but this has limits.
- Third, being able to propose a clean PRs depends on the importance of the project for me of course. I always respect projects and the time that have put on and try to follow what rules they have but this comes last. FYI I have made changes  to my VS to match your settings and still these changes you mentioned got out of hand. In the end this was the best i could do in the time I had for it.
- Forth and last, It is just a PR and you got the idea, if you don't like the code, reject it and implement it yourself. It doesn't really matter.

knocte: > Second, reformatting/resetting/reverting/etc takes more time for me than you as the owner

You're completely wrong here, because you already have your branch and you just need to push new commits to it, but for me it's more annoying to have to recover your branch, create my own one making sure I keep the authorship of the commit, etc.

>  I don't have the environment and the tools/settings as you do

What environment/tools/settings do I have compared to you? I probably use the same IDE as you, or a different one, but this is not related at all with the maintainership of this project (I'm actually just a contributor, so I have limited knowledge about the project, I'm just helping sochix to review PRs).

> Forth and last, It is just a PR and you got the idea, if you don't like the code, reject it and
> implement it yourself. It doesn't really matter.

It's actually code that seems to be not crossplatform, so I actually think it's not a good idea to merge it.

You have a good option on your side, though, to not need to maintain a fork of TLSharp: use the optional delegate that can be provided since this change: https://github.com/sochix/TLSharp/commit/48077961ae730f84f1624f131dd42eb70db18cb6

knocte:
Closing as per last comment.

hosseinGanjyar:
Have this page relation with this issue?
https://github.com/sochix/TLSharp/issues/10

PR #548, opened 2017-08-12
Fix issue with leaving session in a invalid state if GetFile throws an exception after reconnecting to another DC

MichaelManley:
In TelegramClient.GetFile, if the file is in another DC, it reconnects to the DC to be able to get the file. In an exception is thrown after that whilst trying to download the file it doesn't reconnect to the original DC which leaves the session in an invalid state.

This wraps the calls after the ReconnectToDCAs in a try finally to ensure that whatever happens we always try to do the reconnect to the original DC.


knocte:
@sochix can you review?

PR #566, opened 2017-09-07
Added two methods - GetUserProfileImage and GetUserPhotos

aslobodyanuk:
Had to write this for my project and now i suggest you adding them here. You may want to check catch blocks in my methods to ensure they are correct. There is quite a mess in GetUserProfileImage method but it seems to work properly, so that i decided to suggest it not matter what.

### DIFF
--- a/TLSharp.Core/TelegramClient.cs
+++ b/TLSharp.Core/TelegramClient.cs
@@ -347,6 +347,110 @@ namespace TLSharp.Core
 
             return result;
         }
+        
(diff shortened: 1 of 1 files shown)


knocte:
@aslobodyanuk thanks for your contribution!

It seems you're using different line endings than the original sources, so then the diff commit is quite unreadable. Can you fix this please?

knocte:
@aslobodyanuk ping?

PR #578, opened 2017-09-24
Cleaned Codes and Removed UnUsed Usings

ghost1372:
i cleaned codes and removed Unused Usings


TLSharp.Core/MTProto/Crypto/AES.cs
```diff
@@ -216,17 +215,15 @@ public static byte[] XOR(byte[] buffer1, byte[] buffer2)
...
     }
 
-
     // AES engine implementation
 
     public class AesEngine
     {
         // The S box
         private const uint m1 = 0x80808080;
+
```
knocte: why this new line?

TLSharp.Core/MTProto/Crypto/AES.cs
```diff
@@ -576,7 +573,8 @@ public int ProcessBlock(byte[] input, int inOff, byte[] output, int outOff)
             {
                 EncryptBlock(WorkingKey);
             }
-            else {
+            else
+            {
```
knocte: why this change?

TLSharp.Core/Auth/Step3_CompleteDHExchange.cs
```diff
@@ -206,4 +203,4 @@ public Step3_Response FromBytes(byte[] response)
             }
         }
     }
-}
+}
```
knocte: why are you changing all the ending of files?

ghost1372: i use CodeMaid Cleaner for cleaning code so this is just for increase the readability of the code

aarani: i think there is a newline difference here 🤔🤔

ghost1372: no new  code, just some usings removed (that dont used) and white spaces removed.

aarani:
Anyway this pull request cannot get merged as entire TeleSharp.TL is generated so every time I generate new layer they will replace and create not necessary git diff

knocte:
@ghost1372 can you separate this PR in two different commits: one that removes unused usings, and one that cleans up the code (but that doesn't touch autogenerated files, like @aarani says)?

knocte:
There are merge conflicts now. If you can rebase and address the feedback requested, please reopen the PR, thanks.

PR #581, opened 2017-09-26
Updating JSON link to last schema

munrocket:
Change latest scheme in JSON format to aarani tl-schema-66 file


MasihAkbari:
@knocte  i think this PR can be merged

knocte:
@MasihAkbari I would prefer to merge a PR that tells anyone how to generate that JSON file instead of depending on that file to be there. Imagine @aarani's account gets closed or something, then we would lose this info.

knocte:
Superseded by https://github.com/sochix/TLSharp/pull/906

PR #610, opened 2017-10-19
await client.SendCodeRequestAsync Null reference expection 

TERMI80:


knocte:
What is this man? C# files have .cs extension, and need to be grouped in a C# project.

Issue #599, opened 2017-10-07
Unhandled UserMigrationException


arayeji:
TLSharp.Core.Network.UserMigrationException: User located on a different DC: 4. See: https://github.com/sochix/TLSharp#i-get-a-xxxmigrationexception-or-a-migrate_x-error
   at TLSharp.Core.Network.MtProtoSender.HandleRpcResult(UInt64 messageId, Int32 sequence, BinaryReader messageReader, TLMethod request)
   at TLSharp.Core.Network.MtProtoSender.processMessage(UInt64 messageId, Int32 sequence, BinaryReader messageReader, TLMethod request)
   at TLSharp.Core.Network.MtProtoSender.<Receive>d__9.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   at TLSharp.Core.TelegramClient.<SendRequestAsync>d__18`1.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   at TLSharp.Core.TelegramClient.<GetContactsAsync>d__19.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   at TelegramDesktopClient.Main.<Form1_Load>d__2.MoveNext() in

Grepsy:
Actually it can handle these migrations, it's implemented as an `DataCenterMigrationException` which can be caught like here: https://github.com/sochix/TLSharp/blob/0015cf494bea2edce13be36665f177574edadac1/TLSharp.Core/TelegramClient.cs#L148

[...]

PR #622, opened 2017-11-07
Handle DataCenterMigration exceptions in all methods exposed by TelegramClient

Grepsy:
This handles the data center migration in more cases.

Fixes bugs like mentioned in #599.

### DIFF
--- a/TLSharp.Core/TelegramClient.cs
+++ b/TLSharp.Core/TelegramClient.cs
@@ -95,6 +95,24 @@ namespace TLSharp.Core
             await ConnectAsync(true);
         }
 
+        private async Task RequestWithDcMigration(TLMethod request) 
+        {
+            var completed = false;
+            while(!completed)
+            {
+                try
+                {
+                    await _sender.Send(request);
+                    await _sender.Receive(request);
+                    completed = true;
+                }
+                catch(DataCenterMigrationException e)
+                {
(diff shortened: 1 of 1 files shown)


knocte:
Thanks for the PR @Grepsy! Can you reword the commit message to be exactly the same as the text in the PR please? This way, it will reference issue #599 automatically when merging it.

Grepsy:
Done.

TLSharp.Core/TelegramClient.cs
```diff
@@ -187,12 +187,20 @@ public bool IsUserAuthorized()
...
             return ((TLUser)request.Response.user);
         }
+        
         public async Task<TLPassword> GetPasswordSetting()
         {
             var request = new TLRequestGetPassword();
 
-            await _sender.Send(request);
-            await _sender.Receive(request);
+            try
```
knocte: actually these try-catches you added are not correct; they need a while loop to work; please see how it's done in other similar try-catch blocks.

knocte: No worries, can you now squash your two commits into one please? Easiest way to do it is using `git rebase -i HEAD~2`

Grepsy:
Ok, I think this should be correctly squashed now.

knocte:
Good stuff thanks!

PR #625, opened 2017-11-09
TL-Layer: Update To Layer 71

aarani:
This Layer needed lots of Manual Working to Update so don't confuse if generator didn't work same


TeleSharp.TL/TL/Channels/TLRequestGetAdminLog.cs
```diff
@@ -0,0 +1,77 @@
...
+    {
+        public override int Constructor
+        {
+            get
+            {
+                return 870184064;
+            }
+        }
+
+        public int flags { get; set; }
```
knocte: @aarani do you mind fixing this to be PascalCase? It's not respecting the MS .NET API guidelines.

TeleSharp.TL/TL/Channels/TLRequestGetAdminLog.cs
```diff
@@ -0,0 +1,77 @@
...
+        public TLAbsInputChannel channel { get; set; }
+        public string q { get; set; }
+        public TLChannelAdminLogEventsFilter events_filter { get; set; }
```
knocte: Another glitch: members should be PascalCase with no underscores

aarani:
Ok, So we have to update generator first. We'll check that

TeleSharp.TL/TL/Channels/TLRequestGetAdminLog.cs
aarani: Pascal or Camel Case ?

knocte: camelCase is for Java (lowercase the first letter, uppercase each new word), but for .NET it is PascalCase (uppercase the first letter of each word, even the first one)

knocte: so it should be "Flags" instead of "flags", and "EventsFilter" instead of "events_filter".

knocte: (for method names, properties and public fields; private fields can be left as camelCase)

aarani: #627

PR #628, opened 2017-11-09
Make TLVector Implement IList

aarani:
@knocte It will make codes lot simpler


aarani:
@knocte I have to say these last 2 commit (this and other merged one) will need changes in README

knocte:
Good catch! Can you add those changes in this PR?

aarani:
@knocte Sorry i have no time to do that but you can easily use TLSharpTests file to do changes

knocte:
Ok, will do next week.

PR #629, opened 2017-11-09
Handle Import & Export Authorization While Reconnecting

### LINKED ISSUE #624: AUTH_KEY_UNREGISTERED thrown after some period (requires new token), if using GetFile
I've been trying to run TLSharp continuously but seem to periodically (almost every day) run into the `AUTH_KEY_UNREGISTERED` error message. Unfortunately I can't just reconnect, because the same error keeps happening instantly.

I've found the only way to fix it is to delete the `session.dat` and re-request the authorization code, which requires manual intervention.

Does anyone else experience this?

aarani:
it can fix #624 (maybe? 🤔) and reconnecting to DCs authenticated (for example:Migrating for Files) and etc

### DIFF
--- a/TLSharp.Core/TelegramClient.cs
+++ b/TLSharp.Core/TelegramClient.cs
@@ -86,6 +86,13 @@ namespace TLSharp.Core
             if (dcOptions == null || !dcOptions.Any())
                 throw new InvalidOperationException($"Can't reconnect. Establish initial connection first.");
 
+            TLExportedAuthorization exported = null;
+            if (_session.TLUser != null)
+            {
+                TLRequestExportAuthorization exportAuthorization = new TLRequestExportAuthorization() { DcId = dcId };
+                exported = await SendRequestAsync<TLExportedAuthorization>(exportAuthorization);
+            }
+
             var dc = dcOptions.First(d => d.Id == dcId);
 
             _transport = new TcpTransport(dc.IpAddress, dc.Port, _handler);
@@ -93,6 +100,13 @@ namespace TLSharp.Core
             _session.Port = dc.Port;
 
             await ConnectAsync(true);
+
+            if (_session.TLUser != null)
+            {
+                TLRequestImportAuthorization importAuthorization = new TLRequestImportAuthorization() { Id = exported.Id, Bytes = exported.Bytes };
+                var imported = await SendRequestAsync<TLAuthorization>(importAuthorization);
+                OnUserAuthenticated(((TLUser)imported.User));
+            }
         }
 
         private async Task RequestWithDcMigration(TLMethod request) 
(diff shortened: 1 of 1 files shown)


knocte:
I guess this is the intended fix for #624 , can you include this number in the commit message please?

aarani:
It's not fix for that, It will fix reconnecting to different DCs (after Authentication)

aarani:
As we always use reconnecting to different DCs before SendCode , We didn't notice problem before

aarani:
but if you want to migrate to download a file you will have problem

knocte:
> It's not fix for that

Still it's useful to note that "it can fix #624 and other issues".

aarani:
Now I'm seeing it's integrated inside exception handler (totally incorrect place) : https://github.com/sochix/TLSharp/blob/master/TLSharp.Core/TelegramClient.cs#L300
and Developer who wrote that totally forgot to save new authentication key in session. i suggest you to Partial-revert commit that wrote it

aarani:
Ok Git Blame shows that it's Ilya Commit with hash (b06f8a8e1174f0d2580f36039054adb25d62403e) that cause problem here

knocte:
@aarani can you add a fix for that?

aarani:
@knocte Unfortunatly I have no idea what he did

Grepsy:
It seems to me @aarani fix is the correct one, and the exception handling in `GetFile` from Ilya will never be called any more with this fix. Flow:

1. `GetFile` calls `SendRequestAsync`
2.  which calls `RequestWithDcMigration`
3.  which catches all `DataCenterMigrationException`s (which `FileMigrationException` inherits)
4.  which in turn calls `ReconnectToDc` with @aarani's export/import code

This means the `catch (FileMigrationException ex)` is never executed.

Can you guys verify?

knocte:
I don't have time at the moment (maybe next week), but @aarani if you can confirm @Grepsy's findings, go ahead and just remove GetFile's catch(FileMigrationException) block in a 2nd commit of this PR, and we're done here.

knocte:
@aarani I saw the push of your last commit, does this mean that you confirm what @Grepsy said?

aarani:
@knocte I think he’s right, Because of 3ba3ea53fde4a61fdeed01f3b75d2db8007875f7 All MigrationException will handle by SendRequest itself and that catch phrase in GetFile is useless

knocte:
@aarani your second commit includes lots of unneeded whitespace changes

Grepsy:
I've been running this code for the last weekend and didn't have any issues. No guarantees though.

aarani:
@knocte Fixed

knocte:
Ok, this PR looks good diff-wise. I would like to have slightly better commit messages though:

1. On the first commit, please include the text `closes #624`, this way I think github will close the issue and reference it from the commit, when we merge the PR.
2. On the second commit, please state why it's old code. For example simply stating that the catch block could never be reached, as @Grepsy said.

aarani:
@knocte Done

knocte:
good stuff!!

PR #630, opened 2017-11-09
Propose new way to support project

aarani:
I want to suggest slack as it's lot easier,faster,more manageable than gitter to support project.


aarani:
@knocte i suggested using gitter a year ago but as using gitter is so hard because it doesn't have mobile app i suggest using slack instead now.

knocte:
Haha, but man, this is a telegram library. We had gitter because it integrates well with github, but if we move away from gitter, ideally we should make a telegram group.

aarani:
@knocte Telegram group is ok but there are problems :
1- Public Groups not working on Spam-Reported Countries like India & Iran 
2- Private Groups is not manageable at all so it will have chaos after sometimes but i still can create group managing bot, but it's still doesn't have categorizing or other things slack have
---
P.S : slack & telegram are seperated things as they're purpose are totally different

knocte:
@aarani: I guess you have read this: https://news.ycombinator.com/item?id=16539857 . That's the last straw, in my opinion, to never use slack ever again. It's a walledgarden, in the same way whatsapp is. The majority of people that get involved with the telegram technology is because they want to fight walled gardens like whatsapp, so I think it would be stupid to have a slack channel about TLSharp.

I've added links to gitter and telegram channels in the README now: https://github.com/sochix/TLSharp/commit/c1ce5e25baac5d05d4a447c5c563d7a063b4dd14 Feel free to join us there!

PR #717, opened 2018-03-02
Added connection checks

ferferga:
Added connection checks, in order to see if the user is connected to the internet or not


knocte:
This approach is wrong. You cannot put a giant try-catch, filtering by the "Exception" class type (which means it will catch every kind of exception, not only the one caused by an offline connection), and be done with it. Also, throwing non-specific exception types is a bad coding practice.

PR #728, opened 2018-03-19
update README.md to support new PascalCase changes

MasihAkbari:
update README.md to support new PascalCase changes of API and some other fixes

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,13 +81,13 @@ You can call any method on authenticated user. For example, let's send message t
   var result = await client.GetContactsAsync();
 
   //find recipient in contacts
-  var user = result.Users.lists
+  var user = result.Users
 	  .Where(x => x.GetType() == typeof (TLUser))
 	  .Cast<TLUser>()
-	  .FirstOrDefault(x => x.phone == "<recipient_phone>");
+	  .FirstOrDefault(x => x.Phone == "<recipient_phone>");
 	
   //send message
-  await client.SendMessageAsync(new TLInputPeerUser() {user_id = user.id}, "OUR_MESSAGE");
+  await client.SendMessageAsync(new TLInputPeerUser() {UserId = user.Id}, "OUR_MESSAGE");
 ```
 
 Full code you can see at [SendMessage test](https://github.com/sochix/TLSharp/blob/master/TLSharp.Tests/TLSharpTests.cs#L87)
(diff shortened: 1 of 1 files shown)


knocte:
Thanks for your contribution! I'll make a quick review:

README.md
```diff
@@ -81,30 +81,30 @@ You can call any method on authenticated user. For example, let's send message t
   var result = await client.GetContactsAsync();
 
   //find recipient in contacts
-  var user = result.Users.lists
+  var user = result.Users
```
knocte: did you miss to write .Lists here?

README.md
```diff
@@ -81,30 +81,30 @@ You can call any method on authenticated user. For example, let's send message t
...
+  await client.SendMessageAsync(new TLInputPeerUser() {UserId = user.Id}, "OUR_MESSAGE");
 ```
 
 Full code you can see at [SendMessage test](https://github.com/sochix/TLSharp/blob/master/TLSharp.Tests/TLSharpTests.cs#L87)
 
 To send message to channel you could use the following code:
 ```csharp
   //get user dialogs
-  var dialogs = await client.GetUserDialogsAsync();
+  var dialogs = (TLDialogs) await client.GetUserDialogsAsync();
```
knocte: this change seems unrelated to PascalCase API?

README.md
```diff
@@ -81,30 +81,30 @@ You can call any method on authenticated user. For example, let's send message t
...
 
   //find channel by title
-  var chat = dialogs.chats.lists
+  var chat = dialogs.Chats
```
knocte: same here about missing .Lists

README.md
```diff
@@ -169,12 +169,12 @@ Don't panic. You can call any method with help of `SendRequestAsync` function. F
...
   {
-    action = new TLSendMessageTypingAction(),
-    peer = peer
+    Action = new TLSendMessageTypingAction(),
+    Peer = new TLInputPeerUser() { UserId = user.Id }
   };
 
   //run request, and deserialize response to Boolean
-  return await SendRequestAsync<Boolean>(req);
+  return await client.SendRequestAsync<Boolean>(req);
```
knocte: mmm, but `client` variable is not declared anywhere

README.md
MasihAkbari: no, there is no need for it and having it cause compile error

README.md
MasihAkbari: no, but if you remove this cast the "dialogs.Chats" gets compile error

README.md
MasihAkbari: and same reason here, there is no need for it and having it cause compile error

README.md
knocte: ok good to know

README.md
knocte: I'm not talking about dialogs.Chats here

README.md
MasihAkbari: well you can't call SendRequestAsync without it's client, client is defined at the top of README file

knocte: good point!

README.md
MasihAkbari: i edit readme file so a user can copy and paste snippets into their project and successfully compile it, if this cast wheren't here code cause compile error

knocte: what would be the compiler error here?

README.md
knocte: ok

README.md
MasihAkbari: `GetUserDialogsAsync` function returns `TLAbsDialogs` object and is a parent class that `TLDialogs` inheritance from it. `TLAbsDialogs` dosen't have a member called `Chats`, so you should cast return value of `GetUserDialogsAsync` to child class of `TLDialogs` that dose have a member called `Chats`

knocte: makes complete sense, thanks for your patience replying to my questions! merging now

PR #732, opened 2018-03-20
fix compilation error for getting dialogs in README.md file

MasihAkbari:
i missed this in my previews PR, this is the correct code

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,7 +95,7 @@ Full code you can see at [SendMessage test](https://github.com/sochix/TLSharp/bl
 To send message to channel you could use the following code:
 ```csharp
   //get user dialogs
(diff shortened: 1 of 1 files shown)


MasihAkbari:
I'm very puzzled that in the library test code this line is wrong and should throw exception like this [image](https://pasteboard.co/HcPTbkx.jpg)

this is a link to test code : https://github.com/sochix/TLSharp/blob/master/TLSharp.Tests/TLSharpTests.cs#L193

MasihAkbari:
looking at appveyor details it looks like test are just compiled and not run

knocte:
> looking at appveyor details it looks like test are just compiled and not run

Wow! we should fix this! If you have time to look at it? Good catch!

PR #747, opened 2018-04-12
Remove Duplicate Code

aarani:
Serializer now write constructor 2 times and it makes API request fail


aarani:
P.S : it only happen when you use "byte[] Serialize()"

knocte:
> it makes API request fail

You mean it fails because of your commit that changed from GetBuffer to ToArray?

aarani:
no It’s not related to other PR if You check SerializeBody function it write constructor too so we write it two times now and API return error in that case

PR #687, opened 2018-01-22
[Core] Change libs to netstandard

LORDofDOOM:
Make it run on Linux, rename unused files to *.cs.bak

PR #763, opened 2018-05-17
Upgraded to netstandard.

guylando:


seriouz:
Would love to see this getting into the main repo!

TLSharp.Tests.NUnit.old/TLSharp.Tests.NUnit.csproj
```diff
@@ -0,0 +1,54 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project DefaultTargets="Build" ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
knocte: why are you adding these files with the .old extension? just delete the old stuff

knocte: Ok, please ask someone to test it, and if it works well, remove the .old files. Then after that we can merge this PR.

guylando: @seriouz can you test it?

LORDofDOOM:
I've already added a proper PR some month ago: https://github.com/sochix/TLSharp/pull/687

As we already discuss in my PR upgrade to 4.6.1 is needed to respect coding rules for netcore/netstandard (net 4.5 is not compatible). Without 4.6.1 it could break on any library upgrade or is not working as library in other projects where coding follows netstandard rules.

Also BigMath and Ionic.Zlib are no netstandard compatible versions (PCL is deprecated by Microsoft for netstandard), for both are netstandard Version available (see my PR)

ramtinak:
@LORDofDOOM for bigmath I've ported to NetStandard a long time ago:
https://www.nuget.org/packages/BigMathNetStandard/

and there is an Ionic.Zlib nestandard port too:
https://www.nuget.org/packages/Iconic.Zlib.Netstandard/

LORDofDOOM:
@ramtinak: I know that, they also in my PR (https://github.com/sochix/TLSharp/pull/687) - I've wrote in **this** PR the versions are wrong (no netstandard)

ilyalatt:
I created [a fork](https://github.com/ilyalatt/TLSharp.Fork) with netstandard and published it via nuget, it works in my project.

knocte:
@ilyalatt can you create a PullRequest please?

PR #770, opened 2018-05-29
issue#768 fix exception 'Constructor Invalid Or Context.Init Not Called !'

VladSnap:
I used my implementation of FileSessionStore so that I could change the path to the session file.
When you load a session (before creating TelegramClient) from a file (if session is authorized), when the TLUser is deserialized, the 'Constructor Invalid Or Context.Init Not Called !' (inner NullReferenceException) error is thrown.

NullReferenceException here
screenshot_70

It is necessary to initialize TLContext as in the commit and the problem will go away.


knocte:
Looks good, thanks!

Issue #827, opened 2018-12-17
Layer 82 or newer


nimix3:
Hello,

Anyone has any opinion to update to newer (than 66) API layer?!
TLGenerator with newer layer schema cannot create correct class files or entire core need to update.

Good Luck.

nimix3:
Ok, let's do it guys,

I try on Layer 74 but unsuccessful, everything compiled successfully but in send and receive of packets cannot cast correctly, I think some part of the core need to regenerate.

is there anyone try to update layers?

nimix3:
[TeleSharp.TL.zip](https://github.com/sochix/TLSharp/files/2685320/TeleSharp.TL.zip)

The file of TL part attached...

nimix3:
> > TLGenerator with newer layer schema cannot create correct class files
> 
> Why not?

You can check it by your self, newer layers has some missing class / method after conversion by TLGenerator!

I attach the file above just compile it together with TLSharp core and try to register a phone number.

knocte:
> You can check it by your self, newer layers has some missing class / method after conversion by TLGenerator!

How can I check it by myself? Please be more specific, tell us what you did and the error you get.

nimix3:
> > You can check it by your self, newer layers has some missing class / method after conversion by TLGenerator!
> 
> How can I check it by myself? Please be more specific, tell us what you did and the error you get.

Ok, First of All, download the [tl-schema.json](https://github.com/sochix/TLSharp/files/2690454/tl-schema.zip) file then copy in TLGenerator debug folder  "...\TeleSharp.Generator\bin\Debug" then run the TeleSharp.Generator.exe file and wait, after some seconds a new folder created with the name "TL" that there are many *.cs file inside of this folder!

Copy the TL folder and paste in "TeleSharp.TL" main folder then open the solution and compile the project, in the end, you have two files "TeleSharp.TL.dll" and "TLSharp.Core.dll" in "...\TLSharp.Core\bin\Debug" folder now you can use them inside your own projects.

when you want to use them you will get an error when trying to Register a new number, exactly in Connect part, the serialized body received by telegram cannot be deserialized and cause errors!

Please advise if you compile a newer API layer that 66 that is embedded in this repository! if not please be active and just do it, I bet you it takes few minutes not more!

knocte:
> I bet you it takes few minutes not more!

I've never done this myself, so not sure what you're talking about. I just helped here and there, but never tinkered with the Generator.

[...]

Issue #696, opened 2018-01-31
How all dialogs, more 100 dialogs?


InokentiyVas:
Please write how to get all the dialogs. I can not understand offset:
```
public async Task<TLAbsDialogs> GetUserDialogsAsync()
{
	var peer = new TLInputPeerSelf();
	return await SendRequestAsync<TLAbsDialogs>(
		new TLRequestGetDialogs() { OffsetDate = 0, OffsetPeer = peer, Limit = 100 });
}
```

lyamin259:
Hi, i have same problem... Have you been able to solve this problem?

mHCio:
Check this out: #532

hosseinGanjyar:
Did you find any solution?

hosseinGanjyar:
@InokentiyVas Please show a sample code, if you found a solution.
Thanks

theFra985:
As a refernce I used the `offsetDate` parameter

```cs
var lastDate = 0;
while (true) {
    var dialogs = await client.GetUserDialogsAsync(lastDate);
    foreach (var chat in dialogs is TLDialogs d ? d.Chats : ((TLDialogsSlice) dialogs).Chats) {
        switch (chat) {
            case TLChannelForbidden _:
            case TLChatForbidden _:
                break;
            case TLChat c:
                lastDate = c.Date;
                break;
            case TLChannel c:
                lastDate = c.Date;
                break;
            default:
                throw new NotSupportedException($"Unsupported chat type {chat.GetType()}");
        }
    }
    if (dialogs.Dialogs.Count < 100) // Not sure but it seem to work
        break;
}
```

If you want to compact the code it will become
```cs
TLAbsDialogs dialogs = null;
while (true)
	dialogs = await client.GetUserDialogsAsync(
		(dialogs is TLDialogs d ? d.Chats : ((TLDialogsSlice) dialogs)?.Chats)
		?.Select(chat => chat is TLChat c ? c.Date : chat is TLChannel chan ? chan.Date : -1)
		.Last(i => i >= 0) ?? 0);
```

PR #831, opened 2019-01-04
Fixed OffSet paramets for GetUserDialogsAsync & GetHistoryAsync

mkellerman:
Added missing OffSet parameters.


knocte:
Question: were this paramaters missing because we're behing the layer version? AFAIU, we're using Layer 66. But according to this (https://github.com/sochix/TLSharp/issues/827) we may be lagging behind. Could you confirm? More info: https://stackoverflow.com/questions/41691205/telegram-latest-layer

knocte:
(Note to self: this would fix this bug: https://github.com/sochix/TLSharp/issues/696 )

PR #835, opened 2019-01-17
add more telegram server public key

nimix3:
add more telegram server public key


knocte:
Where did you get this from?

nimix3:
Telegram has many server rsa public key, I get some tested and properly from "telethon" a python telegram client api project and convert PEM format to ASN1 then decode/encode for suitable format for TLSharp and add to this project. As I test it everything seems ok and working...

knocte:
Isn't there an official telegram's webpage where these keys are?

nimix3:
It generated by telegram and can be decrypted on telegram servers only as you know.
Telegram does not have any public page for RSA keys for everyone you can get it from Developers part of telegram website, I can't understand your exact mean!

knocte:
> It generated by telegram and can be decrypted on telegram servers only as you know.

Why would I know that?

>  you can get it from Developers part of telegram website

where is that Developers part? please mention the URL so I can compare

knocte:
You have to understand I cannot accept any random gibberish that any person from the internet wants to put in the repo, it needs to have an explanation.

Issue #719, opened 2018-03-11
DC Migration loop


iz80:
Scenario:

0. Create a client.
1. Try to get you contact list with an authenticated account.
2. Try downloading profile image of another user.
3. Try accessing your contact list again (just an example).

Between step 0 and 1:
- DC changes from 3 (default) to 4 (where the authenticated user is located)

Between step 1 and 2:
- Get a "FILE_MIGRATE_X" where dc_id is 2
- TLSharp handles the DC Migration and downloads the file
- TLSharp saves the new dc 2 authentication in session

On step 3:
- Get a "USER_MIGRATE_X" where dc_id is 4
- TLSharp tries to handle DC migration from 2 to 4
- On TLRequestExportAuthorization, "USER_MIGRATE_X" (dc_id 4) happens again
- Here the loop starts and never ends or until telegram server breaks the connection.

Maybe the right solution is to connect to the DC where the file is located without replacing the authentication in the session, but just to use this authentication for the sake of downloading the file. Note that when we get the file object, we already know the DC it's located on.

knocte:
@iz80 great analysis!

> Maybe the right solution is to connect to the DC where the file is located without replacing the authentication in the session, but just to use this authentication for the sake of downloading the file. 

Sounds like a sound approach. Do you mind trying to implement it, and if it works for you, propose it as a pull request? Thanks!

iz80:
@knocte I would love to, but I'm short on time and afraid that committing to this feature is not something I could do right now. If I could, I would have probably opened this issue with a pull request already.

MahdiHosseiniOfficial:
I have the same problem

MahdiHosseiniOfficial:
@knocte  hi , Did you find a solution to this problem?

knocte:
@MahdiHosseiniOfficial if I had found a solution, I would have already fixed it!

camanchaca:
Got the same problem after sequental 10-20 images download. New authorization - and again 10-20 images. Any ideas?

MahdiHosseiniOfficial:
I'm in the join groups or forward or send message and another method This problem is ready for me
But I have an idea . I'm testing it . 
If the problem is resolved . I'm solve the problem with pull requests

Jarzamendia:
Any news about that? I'm research a method to avoid too.

guyho2:
Anyone solved it somehow???
After calling GetFile few times the session is corrupted.

camanchaca:
[...]

PR #838, opened 2019-01-28
Fix dc loop #719

ArmeF97:
Inserting a counter, after "times", throws an exception, that should be handled

### DIFF
--- a/TLSharp.Core/Auth/Authenticator.cs
+++ b/TLSharp.Core/Auth/Authenticator.cs
@@ -5,31 +5,31 @@ namespace TLSharp.Core.Auth
 {
     public static class Authenticator
     {
-        public static async Task<Step3_Response> DoAuthentication(TcpTransport transport)
+        public static Step3_Response DoAuthentication(TcpTransport transport)
         {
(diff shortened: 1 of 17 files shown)


ArmeF97:
Maybe it's a fix for #719

maurodx:
Probably this PR can't be accepted because you change all the async-await and also .NET Framework version, methods name, etc...

ArmeF97:
> 
> 
> Probably this PR can't be accepted because you change all the async-await and also .NET Framework version, methods name, etc...

I'm sorry, I don't know how github works... I did the pull request with the first commit, than I did some commits to my fork, as a test, I didn't know that they were going to be added to this pr

ArmeF97:
Should I do a new branch with just that commit?

knocte:
@ArmeF97 sorry for the late reply, but I committed a different approach in master branch, see here: 60a3c6235792163e9c74e9c24b1ca7800a220ce2

ArmeF97:
@knocte no problem. I'm glad I was of help. I'm sorry I didn't know how github works in pr... 

Now I now more and can help in future! 👍

Issue #773, opened 2018-06-04
Exception when reconnect to DC "An address incompatible with the requested protocol was used"


huyvu88:
I got this error message when TLSharp try to reconnect to DC once DataCenterMigrationException appears:
"An address incompatible with the requested protocol was used [2001:b28:f23f:f005::a]:443"

Take a look at https://github.com/sochix/TLSharp/blob/master/TLSharp.Core/TelegramClient.cs - RequestWithDcMigration() catch block. 

I google this error and found out it should be the mismatch between IpV6 and IpV4 which lead me to https://github.com/sochix/TLSharp/blob/master/TLSharp.Core/Network/TcpTransport.cs 
` public TcpTransport(string address, int port, TcpClientConnectionHandler handler = null)
        {
            if (handler == null)
            {
                _tcpClient = new TcpClient();

                var ipAddress = IPAddress.Parse(address);
                _tcpClient.Connect(ipAddress, port);
            }
            else
                _tcpClient = handler(address, port);
        }
`
I believed that some returned DC which uses IpV6 that cause this error, please help to fix it

huyvu88:
@ctphu I have a workaround while waiting @sochix have time to take a look.
datacenter options (dcOptions variable) returned in ConnectAsync() method has a couple of ipv6 ver and ipv4 ver (example: dc's id 5 could has 2 public ip (one is ipv6 and the other is ipv4), you can modify the TLSharp code that get the DC by dc's id to get ipv4 only then it worked, i tried,

ctphu:
@huyvu88 Can you give your change please :) i will buy a coffee for you if you in Sai Gon

huyvu88:
![image](https://user-images.githubusercontent.com/30425589/41193846-a1cfff74-6c3c-11e8-979d-797f18200ae1.png)
find this line of code in TelegramClient.cs, you're welcome.  I live in Hanoi ... ship me a coffee if it worked to you LOL

ctphu:
Many thanks 👍 it work for me. I think you can close this issue.

knocte:
I'm reopening the issue because no one has merged a fix in TLSharp. @huyvu88 can you provide your fix in a diff/PR form instead of a screenshot please?

huyvu88:
Hi @knocte, this could be a workaround - find this line of code in TelegramClient.cs
`var dc = dcOptions.First(d => d.Id == dcId);`
Modify it as
[...]

PR #853, opened 2019-06-19
Applied fix for sochix/TLSharp#773

CloneDeath:
Ignored IPv6 datacenters, since it's currently not supported.
Added .gitignore for Jetbrains Rider IDE files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -178,6 +178,7 @@ ClientBin/
 *.publishsettings
 node_modules/
 orleans.codegen.cs
+.idea/
 
 # RIA/Silverlight projects
 Generated_Code/
--- a/TLSharp.Core/Network/TcpTransport.cs
+++ b/TLSharp.Core/Network/TcpTransport.cs
@@ -16,10 +16,11 @@ namespace TLSharp.Core.Network
         {
             if (handler == null)
             {
-                _tcpClient = new TcpClient();
-
                 var ipAddress = IPAddress.Parse(address);
-                _tcpClient.Connect(ipAddress, port);
+                var endpoint = new IPEndPoint(ipAddress, port);
+
+                _tcpClient = new TcpClient(endpoint);
+                _tcpClient.Connect(endpoint);
             }
             else
                 _tcpClient = handler(address, port);
--- a/TLSharp.Core/TelegramClient.cs
+++ b/TLSharp.Core/TelegramClient.cs
@@ -89,7 +89,7 @@ namespace TLSharp.Core
                 exported = await SendRequestAsync<TLExportedAuthorization>(exportAuthorization);
             }
 
-            var dc = dcOptions.First(d => d.Id == dcId);
+            var dc = dcOptions.First(d => d.Id == dcId && !d.Ipv6);
             var dataCenter = new DataCenter (dcId, dc.IpAddress, dc.Port);
 
             _transport = new TcpTransport(dc.IpAddress, dc.Port, _handler);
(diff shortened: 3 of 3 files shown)


knocte:
> Ignored IPv6 datacenters, since it's currently not supported.

Not supported? Where did you get that from?

CloneDeath:
Take a look at #773 

"An address incompatible with the requested protocol was used [2001:b28:f23f:f005::a]:443"

This happens whenever trying to connect to a datacenter with an IPv6 address. I can confirm that this has halted the incompatible protocol exceptions (by avoiding using them and just using IPv4 datacenters instead).

It is true that it doesn't fix the problem, just bypasses it. If IPv6 support is ever added, then we would need to filter out that small `IsIpV6` check. In the meantime, at least it won't crash with no workaround anymore.

knocte:
> It is true that it doesn't fix the problem

Exactly.

> If IPv6 support is ever added

What makes you think that IPv6 is not already supported?

CloneDeath:
"An **address** incompatible with the requested protocol was used **[2001:b28:f23f:f005::a]:443**"

The exception being thrown whenever trying to connect to an IPv6 address is what makes me think that it's not already supported.

knocte:
> The exception being thrown whenever trying to connect to an IPv6 address is what makes me think that it's not already supported.

Well, that might just be because the origin is IPv4.

CloneDeath:
Thanks for the tip. I'm looking into it more now...
It looks like the TcpClient default constructor only supports IPv4, which we use, and *could* be the source of the bug where it looks like IPv4 isn't working. I'm testing it now...

Constructor only works for IPv4:
https://docs.microsoft.com/en-us/dotnet/api/system.net.sockets.tcpclient.-ctor?view=netframework-4.8#System_Net_Sockets_TcpClient__ctor

Where we are using IPv4-only constructor:
https://github.com/sochix/TLSharp/blob/master/TLSharp.Core/Network/TcpTransport.cs#L19

P.S. I appreciate your patience and thoroughness!

CloneDeath:
Yup. The TcpClient constructor was the problem. I can connect to IPv6 addresses now.

You can reproduce the issue & fix simply with this snippet of code (works with any IPv6 address... the broken case always fails, even with the IPv6 loopback address, throwing the same address exception).

```
using System;
using System.Net;
using System.Net.Sockets;

namespace IPv6Test {
	class Program {
		static void Main() {
			var ip_address = "::1";
			try {
				ReproduceIssue(ip_address);
				Console.WriteLine("Current Method [Passed]");
			}
			catch (SocketException ex){
				Console.WriteLine("Current Method [Failed]: " + ex.SocketErrorCode);
			}

			try {
				FixedIssue(ip_address);
				Console.WriteLine("Fix [Passed]");
			}
			catch (SocketException ex){
				Console.WriteLine("Fix [Failed] " + ex.SocketErrorCode);
			}
		}

		static void ReproduceIssue(string testAddress) {
			var ipAddress = IPAddress.Parse(testAddress);

			var client = new TcpClient();
			client.Connect(ipAddress, 443);
		}

		static void FixedIssue(string testAddress) {
			var ipAddress = IPAddress.Parse(testAddress);
			var endpoint = new IPEndPoint(ipAddress, 443);

			var client = new TcpClient(endpoint);
			client.Connect(endpoint);
		}
	}
}
```

Which outputs:

> Current Method [Failed]: AddressFamilyNotSupported
> Fix [Passed]

CloneDeath:
Alright, updated the PR to support IPv6 addresses.

knocte:
And this fixes #773?

BTW can you remove all whitespace changes? you're modifying two other files for no reason, only TLSharp.Core/Network/TcpTransport.cs should be modified in this PR.

CloneDeath:
Sure, closing this PR, and opening a new one for the new branch `fix-ipv6-addresses`

PR #855, opened 2019-06-27
Modified TcpClient setup to work with IPv6 addresses, fixing #773.

CloneDeath:
@knocte He's the new PR, with the whitespace gone. Sorry if this is getting confusing.

This is a continuation from #853. There, I attempted to avoid the crash by skipping IPv6 datacenters. In this branch, I actually fixed the crash by changing the way the TcpClient was setup, allowing IPv6 addresses.


knocte:
Good work!

Issue #862, opened 2019-07-06
The requested address is not valid in its context


WatchDogsDev:
Hello , When I start the new version and try to connect , It throws "the requested address is not valid in its context" socket exception. This issue has appeared since version **0.1.0.390**

My code :

```
 var Client = new TelegramClient(API_ID, API_HASH);
                await Client.ConnectAsync();
                await Task.Delay(250);
                sHash = await Client.SendCodeRequestAsync(number);
```

overw1ite:
same problem

alishokr:
I had the same issue.
i fixed it by making changes to TcpTransport constructor.

```
public TcpTransport(string address, int port, TcpClientConnectionHandler handler = null)
        {
            if (handler == null)
            {
                /*var ipAddress = IPAddress.Parse(address);
                var endpoint = new IPEndPoint(ipAddress, port);*/

                _tcpClient = new TcpClient();
                _tcpClient.Connect(address,port);
            }
            else
                _tcpClient = handler(address, port);
        }
```

knocte:
@alishokr if you do that you're basically reverting the last commit which in theory added support for IPv6. We need a better fix that detects if the IP address is IPv4 or IPv6, and use a different constructor of TcpClient() class depending on that.

alishokr:
> @alishokr if you do that you're basically reverting the last commit which in theory added support for IPv6. We need a better fix that detects if the IP address is IPv4 or IPv6, and use a different constructor of TcpClient() class depending on that.

you're right. i needed quick fix.

eordukaya:
You can see this post #859

knocte:
Fixed in 00330c9bd010945285e56027345d3f37bcf4792a

PR #863, opened 2019-07-11
Fix Issues #862

devMikeUA:
Fix connection by IPv4 address.


knocte:
@CloneDeath can you review?

CloneDeath:
(Sorry for the delay, busy all day)
I don't see any problems with this. I'm not able to run it now (but the CI tests pass), but I don't see any problems with using the Address family (based on what I'm reading in the docs) instead.

I approve for merge.

PR #870, opened 2019-09-12
Added timeout to RequestWithDcMigration

### LINKED ISSUE #869: I think RequestWithDcMigration needs timeout
Forever loop caused, if phone without country code passed to SendCodeRequestAsync, for example.

artygrand:
Fix #869

Im just started with c#, so this can be wrong.


knocte:
> Im just started with c#

Right, there's a way to encode time in .NET: the TimeSpan type. Use that instead of integer.

knocte:
As discussed in #869, this is an unacceptable approach to fix this problem sorry.

PR #874, opened 2019-10-02
Add ability to set directory to store sessions

MojtabaTajik:
Now user can set directory to store sessions in there and not inside program's executable

### DIFF
--- a/TLSharp.Core/Session.cs
+++ b/TLSharp.Core/Session.cs
@@ -14,9 +14,18 @@ namespace TLSharp.Core
 
     public class FileSessionStore : ISessionStore
     {
+        private readonly string _storePath;
+
+        public FileSessionStore(string storePath = "")
+        {
+            _storePath = storePath;
+        }
+
         public void Save(Session session)
         {
-            using (var stream = new FileStream($"{session.SessionUserId}.dat", FileMode.OpenOrCreate))
+            string sessionPath = Path.Combine(_storePath, $"{session.SessionUserId}.dat");
+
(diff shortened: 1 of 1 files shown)


TLSharp.Core/Session.cs
```diff
@@ -14,9 +14,18 @@ public interface ISessionStore
 
     public class FileSessionStore : ISessionStore
     {
+        private readonly string _storePath;
+
+        public FileSessionStore(string storePath = "Sessions")
```
knocte: you're providing a default value here which is different than what we had before

MojtabaTajik: The default value added to backward compatibility to previous versions.

knocte: I'm saying that this is not backwards compatible, the "Sessions" string was not there before.

MojtabaTajik: @knocte Now it's compatible.

TLSharp.Core/Session.cs
```diff
@@ -14,9 +14,18 @@ public interface ISessionStore
 
     public class FileSessionStore : ISessionStore
     {
+        private readonly string _storePath;
+
+        public FileSessionStore(string storePath = "")
```
knocte: ok, now I prefer this to have a better type, it should be `DirectoryInfo`, not string. (And it should default to null.)

knocte:
@MojtabaTajik I've finished refining your contribution and pushed the enhancement to master: https://github.com/sochix/TLSharp/commit/e4d0e97571a724c22aa8d519f4e66a1638aaa10e . Thanks!

PR #875, opened 2019-10-02
Add Session readonly field to TelegramClient

MojtabaTajik:
Add Session readonly field which let user read session properties like current logedin user's Username & etc.

### DIFF
--- a/TLSharp.Core/TelegramClient.cs
+++ b/TLSharp.Core/TelegramClient.cs
@@ -29,6 +29,11 @@ namespace TLSharp.Core
         private List<TLDcOption> dcOptions;
         private TcpClientConnectionHandler _handler;
 
+        public Session Session
+        {
+            get { return _session; }
+        }
+
         public TelegramClient(int apiId, string apiHash,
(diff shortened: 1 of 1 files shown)


knocte:
Make it a {get}-only property, not a field, please.

MojtabaTajik:
> Make it a {get}-only property, not a field, please.

It's already get only. The code is expression bodied style of get only.

knocte:
I would prefer to make it explicit just in case (more verbose but more readable).

MojtabaTajik:
@knocte Done

knocte:
Now the git commit message doesn't contain the good description that you had added before: "Add Session readonly field which let user read session properties like current logged-in user's Username & etc.". Please add it.

knocte:
I said the commit message, not the PR title.

knocte:
Do you need help with interactive rebase?

knocte:
@MojtabaTajik I merged your improvement manually in the master branch: https://github.com/sochix/TLSharp/commit/8a15feea7c9357aab5bf6df504d45ae2077009d4

PR #883, opened 2019-10-25
Adds ConfigureAwait(false) and CancellationToken to all async methods

IaRuslan:
Hi, guys.
Thank you for your library. It's very useful. But I've got deadlocks in TcpTransport.Receive, so I added ConfigureAwait(false) to all async method calls to solve the problem. Also I added CancellationToken to methods parameters.

### DIFF
--- a/TLSharp.Core/Auth/Authenticator.cs
+++ b/TLSharp.Core/Auth/Authenticator.cs
@@ -1,35 +1,38 @@
-﻿using System.Threading.Tasks;
+﻿using System.Threading;
+using System.Threading.Tasks;
 using TLSharp.Core.Network;
 
 namespace TLSharp.Core.Auth
 {
     public static class Authenticator
     {
-        public static async Task<Step3_Response> DoAuthentication(TcpTransport transport)
+        public static async Task<Step3_Response> DoAuthentication(TcpTransport transport, CancellationToken token)
         {
+            token.ThrowIfCancellationRequested();
+            
             var sender = new MtProtoPlainSender(transport);
             var step1 = new Step1_PQRequest();
 
-            await sender.Send(step1.ToBytes());
-            var step1Response = step1.FromBytes(await sender.Receive());
+            await sender.Send(step1.ToBytes(), token).ConfigureAwait(false);
+            var step1Response = step1.FromBytes(await sender.Receive(token).ConfigureAwait(false));
 
             var step2 = new Step2_DHExchange();
             await sender.Send(step2.ToBytes(
                 step1Response.Nonce,
                 step1Response.ServerNonce,
                 step1Response.Fingerprints,
-                step1Response.Pq));
+                step1Response.Pq), token).ConfigureAwait(false);
 
-            var step2Response = step2.FromBytes(await sender.Receive());
+            var step2Response = step2.FromBytes(await sender.Receive(token).ConfigureAwait(false));
 
             var step3 = new Step3_CompleteDHExchange();
             await sender.Send(step3.ToBytes(
                 step2Response.Nonce,
                 step2Response.ServerNonce,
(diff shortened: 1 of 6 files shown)


knocte:
Thanks for your contribution! I'm going to do a review now, please bear with me.

TLSharp.Core/Auth/Authenticator.cs
```diff
@@ -1,35 +1,38 @@
...
+﻿using System.Threading;
+using System.Threading.Tasks;
 using TLSharp.Core.Network;
 
 namespace TLSharp.Core.Auth
 {
     public static class Authenticator
     {
-        public static async Task<Step3_Response> DoAuthentication(TcpTransport transport)
+        public static async Task<Step3_Response> DoAuthentication(TcpTransport transport, CancellationToken token)
```
knocte: can you please make CancellationToken parameters be optional, and default to the value `CancellationToken.None` please?

TLSharp.Core/Network/MtProtoPlainSender.cs
```diff
@@ -1,24 +1,27 @@
...
     public class MtProtoPlainSender
     {
-        private int timeOffset;
-        private long lastMessageId;
-        private Random random;
-        private TcpTransport _transport;
+        private int _timeOffset;
+        private long _lastMessageId;
+        private readonly Random _random;
+        private readonly TcpTransport _transport;
```
knocte: thanks for converting some fields to `readonly`! but this is not related to cancellation or deadlock fixing, so can you separate this into a different commit or even different PR?

TLSharp.Core/Network/TcpTransport.cs
```diff
@@ -91,13 +95,7 @@ public async Task<TcpMessage> Receive()
...
         }
 
-        public bool IsConnected
-        {
-            get
-            {
-                return this._tcpClient.Connected;
-            }
-        }
+        public bool IsConnected => _tcpClient.Connected;
```
knocte: please revert this refactoring, it's not related

TLSharp.Core/Network/TcpTransport.cs
```diff
@@ -32,27 +33,29 @@ public TcpTransport(string address, int port, TcpClientConnectionHandler handler
...
+            var tcpMessage = new TcpMessage(_sendCounter, packet);
 
-            await _stream.WriteAsync(tcpMessage.Encode(), 0, tcpMessage.Encode().Length);
-            sendCounter++;
+            await _stream.WriteAsync(tcpMessage.Encode(), 0, tcpMessage.Encode().Length, token).ConfigureAwait(false);
+            _sendCounter++;
         }
 
-        public async Task<TcpMessage> Receive()
+        public async Task<TcpMessage> Receive(CancellationToken token)
```
knocte: this two methods above are public and still don't have default token

knocte: @IaRuslan I've just realized that the same happens in Authenticator.cs, MtProtoPlainSender.cs, and MtProtoSender.cs

knocte: ping

knocte:
I actually looked at the code and the most prevailing style is to *not* have `_` as a prefix for private fields. These are only a few cases, not the majority. So we should rather make it uniform to not have this, instead of adding it to the ones missing it.

knocte:
Hey @IaRuslan , you never pinged me about your last push, so I didn't know this PR was finished. But then there were some other changes which made this PR have conflicts, my bad!

Fortunately @CheshireCaat has picked up the work and brought your changes, and he mentioned you in the commit!: c5a2c816fc4ea920c40d3e0a85708470dced2ac6

Thanks for the work guys!

PR #893, opened 2020-01-09
Port to .NETStandard 2.0

merqlove:
That port contains only nestandard2.0 due crosscompat. NETStandard 2.0 projects can be used under .net.
Dropped .NET only dependencies. Now project uses netstandard deps only.
Removed some old files in Requests, for better compile.
Remove nuspec usage. Now all in .csproj.

If that project need such upgrade merge & change it for project needs.


TeleSharp.Generator/Program.cs
```diff
@@ -426,7 +432,7 @@ public static string WriteReadCode(Param p)
         public static FileStream MakeFile(string path)
         {
             if (!Directory.Exists(Path.GetDirectoryName(path)))
-                Directory.CreateDirectory(Path.GetDirectoryName(path));
+                Directory.CreateDirectory(Path.GetDirectoryName($"{currentDirectory}../../../../{path}"));
```
knocte: what is this? this doesn't seem to be related to .NETStandard change, or it's a changed that is not explained in the commit message

TeleSharp.Generator/TeleSharp.Generator.csproj
```diff
@@ -9,9 +8,15 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>TeleSharp.Generator</RootNamespace>
     <AssemblyName>TeleSharp.Generator</AssemblyName>
-    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
     <FileAlignment>512</FileAlignment>
+    <TargetFrameworks>netcoreapp2.1</TargetFrameworks>
```
knocte: there's no reason to change this project's targetFramework, projects with .NETFramework 4.x can still link to .NETStandard libraries

TeleSharp.Generator/Program.cs
merqlove: I've got error with relative path, so it is `cd` to the project root from bin/debug directory

knocte: what error? this info needs to be in the commit message

PR #896, opened 2020-01-23
Add methods for working with contacts.

CheshireCaat:
Added custom constructor for TLVector. This makes it easy to create a TLVector object from other collections.

Added methods for working with contacts:
- Import contact list
- Delete contact list
- Delete one contact

### DIFF
--- a/TLSharp.Core/TelegramClient.cs
+++ b/TLSharp.Core/TelegramClient.cs
@@ -234,12 +234,42 @@ namespace TLSharp.Core
             return (T)result;
         }
 
+        public async Task<TLImportedContacts> ImportContactsAsync(IReadOnlyList<TLInputPhoneContact> contacts)
+        {
+            if (!IsUserAuthorized())
+                throw new InvalidOperationException("Authorize user first!");
+
+            var req = new TLRequestImportContacts { Contacts = new TLVector<TLInputPhoneContact>(contacts)};
+
+            return await SendRequestAsync<TLImportedContacts>(req);
+        }
+
+        public async Task<bool> DeleteContactsAsync(IReadOnlyList<TLAbsInputUser> users)
+        {
+            if (!IsUserAuthorized())
+                throw new InvalidOperationException("Authorize user first!");
+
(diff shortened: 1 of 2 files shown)


TeleSharp.TL/TLVector.cs
```diff
@@ -11,7 +11,17 @@ namespace TeleSharp.TL
...
+
+        public TLVector(IEnumerable<T> collection)
+        {
+            lists = new List<T>(collection);
+        }
+
+        public TLVector()
+        {
+            lists = new List<T>();
+        }
```
knocte: thanks for your contribution! can you remove this from the PR? seems unrelated with the problem at hand

CheshireCaat: Okay, i'll solve it a bit later

knocte: wait, I think I'm mistaken, sorry, it is related

TLSharp.Core/TelegramClient.cs
```diff
@@ -234,12 +234,42 @@ public async Task<T> SendRequestAsync<T>(TLMethod methodToExecute)
...
+            return await SendRequestAsync<TLLink>(req);
+        }
+
         public async Task<TLContacts> GetContactsAsync()
         {
             if (!IsUserAuthorized())
                 throw new InvalidOperationException("Authorize user first!");
 
-            var req = new TLRequestGetContacts() { Hash = "" };
+            var req = new TLRequestGetContacts {Hash = ""};
```
knocte: what is not related is this change

TeleSharp.TL/TLVector.cs
CheshireCaat: Well, I thought that you want to leave the default constructor empty and leave the list initialization as it was before. If there are no problems, this is very good)

TLSharp.Core/TelegramClient.cs
CheshireCaat: Fixed

PR #899, opened 2020-01-24
Reformat exceptions

CheshireCaat:
Exceptions moved to separate folders:
- Exceptions from MtProtoSender.cs to TLSharp.Core/Network/Exceptions
- Exceptions from TelegramClient.cs to TLSharp.Core/Exceptions


knocte:
This is great thanks! There's only one nitpick: the great text that you have put in the description of this PR should also be in the commit message.

CheshireCaat:
I will keep in mind, thanks!

knocte:
Just one more detail for next PR: the first line of the commit message is a "title", so it shouldn't end with `:`.

PR #901, opened 2020-01-26
Code refactoring

CheshireCaat:
This PR contains some commits that contains code refactoring for:

TeleSharp.Generator:
- Removed unused references in Program.cs and replace 'String' to 'string'
- Added JsonProperty attribute for each property in models
- Models moved to separate folders

TLSharp.Core.MTProto.Serializers:
- Fix naming for methods and remove unused method

TLSharp.Core.DataCenter:
- Remove redundant setters for properties


TLSharp.Core/DataCenter.cs
```diff
@@ -1,21 +1,20 @@
-
 namespace TLSharp.Core
 {
     internal class DataCenter
     {
-        internal DataCenter (int? dcId, string address, int port)
+        internal DataCenter(int? dcId, string address, int port)
```
knocte: no need to change this line above

TLSharp.Core/DataCenter.cs
```diff
@@ -1,21 +1,20 @@
...
         {
             DataCenterId = dcId;
             Address = address;
             Port = port;
         }
 
-        internal DataCenter (string address, int port) : this (null, address, port)
+        internal DataCenter(string address, int port) : this(null, address, port)
```
knocte: same here

CheshireCaat: Fixed it

knocte: can you squash the 2nd commit with the 4th commit?

CheshireCaat: Done

knocte:
I guess you tested TLSharp after these changes?

CheshireCaat:
Yes, i always check that library working fine using small console-client. Should I test specific functions? I guess that this PR not contains changes that may be break functionality of TLSharp.

Only thing that TeleSharp.Generator wasn't tested by me.

knocte:
Well, the generator is the most important thing to test here because you used new Json attributes.

CheshireCaat:
Ok, i tested it now with 23 layer from official docs. It works without any exceptions.

PR #903, opened 2020-01-27
Improvements for async methods

CheshireCaat:
Because there is no more actions from author in PR  #883 i decide to create this PR by myself.

@knocte, i tried to do all fixes which you requested from the author.
All async methods now contains CancellationToken with default value (CancellationToken.None)
And one more fix - i detect that SendAuthenticatetRequestAsync wasn't used in some methods, that need auth.
Therefore i changed it and changed private modifier in SendAuthenticatedRequestAsync to internal. It's needed for methods that exists in UploadHelper.cs.

I'm waiting for your review.

### DIFF
--- a/TLSharp.Core/TelegramClient.cs
+++ b/TLSharp.Core/TelegramClient.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Security.Cryptography;
 using System.Text;
+using System.Threading;
 using System.Threading.Tasks;
 using TeleSharp.TL;
 using TeleSharp.TL.Account;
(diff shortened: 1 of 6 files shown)


CheshireCaat:
Don't know why AppVeyor says that build failed - i successfuly builded solution and tested it.

TLSharp.Core/TelegramClient.cs
```diff
@@ -413,15 +413,15 @@ public async Task<TLAbsMessages> GetHistoryAsync(TLAbsInputPeer peer, int offset
         /// <param name="q">User or chat name</param>
         /// <param name="limit">Max result count</param>
         /// <returns></returns>
-        public async Task<TLFound> SearchUserAsync(string q, int limit = 10, CancellationToken token = default)
+        public async Task<TLFound> SearchUserAsync(string q, int limit = 10, CancellationToken token = default(CancellationToken))
```
knocte: cancellation-related changes should be squashed with the first commit

CheshireCaat: Done!

knocte:
Haha but now you put everything in one commit. You should have changed it so that the auth-changes stay separate. Anyway, it's not a big deal, leave it like that, however, what is a big deal is that these changes are based on another PR by another contributor, so you should mention him/her on the commit message.

PR #905, opened 2020-02-01
Fix problem with TLRequestGetStickerSet

raznahan:
When trying to SendRequest with TLRequestGetStickerSet throws an exception saying "cannot cast object of type TeleSharp.TL.TLStickerSet to TeleSharp.TL.Messages.TLStickerSet".

### DIFF
--- a/TeleSharp.TL/TL/Messages/TLStickerSet.cs
+++ b/TeleSharp.TL/TL/Messages/TLStickerSet.cs
@@ -18,7 +18,7 @@ namespace TeleSharp.TL.Messages
             }
         }
 
-        public TLStickerSet Set { get; set; }
+        public TL.TLStickerSet Set { get; set; }
         public TLVector<TLStickerPack> Packs { get; set; }
         public TLVector<TLAbsDocument> Documents { get; set; }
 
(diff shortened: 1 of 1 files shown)


knocte:
@CheshireCaat please review

CheshireCaat:
@raznahan, a small tip for the future - use a short header for commit, and write more below, for example:

Fix type of TLStickerSet Set property (short description)

Change type of Set property from TLStickerSet to TL.TLStickerSet (more additinal info)

CheshireCaat:
@knocte I am afraid that in the future, if you need to update the layer, this problem may appear again. I’ll try to look at the TeleSharp.Generator for problems when generating such classes later.

knocte:
>  use a short header for commit, and write more below,

The commit message should also include the number of the github issue that is fixing.

>  I’ll try to look at the TeleSharp.Generator for problems when generating such classes later.

Sounds good!

raznahan:
> @raznahan, a small tip for the future - use a short header for commit, and write more below, for example:
> 
> Fix type of TLStickerSet Set property (short description)
> 
> Change type of Set property from TLStickerSet to TL.TLStickerSet (more additinal info)

Thanks! I'm kinda new to git, so will try next time for sure.

Issue #777, opened 2018-06-16
msg_seqno too low


hanigamini:
Hello,
I have this error,please help me.

msg_seqno too low (the server has already received a message with a lower msg_id but with either a higher or an equal and odd seqno) : System.InvalidOperationException: msg_seqno too low (the server has already received a message with a lower msg_id but with either a higher or an equal and odd seqno)
   at TLSharp.Core.Network.MtProtoSender.HandleBadMsgNotification(UInt64 messageId, Int32 sequence, BinaryReader messageReader) in TLSharp-master\TLSharp.Core\Network\MtProtoSender.cs:line 389
   at TLSharp.Core.Network.MtProtoSender.processMessage(UInt64 messageId, Int32 sequence, BinaryReader messageReader, TLMethod request) in TLSharp.Core\Network\MtProtoSender.cs:line 200
   at TLSharp.Core.Network.MtProtoSender.<Receive>d__9.MoveNext() in TLSharp.Core\Network\MtProtoSender.cs:line 144
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter1.GetResult()
   at TLSharp.Core.TelegramClient.<ConnectAsync>d__9.MoveNext() in TLSharp.Core\TelegramClient.cs:line 77
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter1.GetResult()
   at HaniYar.TLSharpMethods.<ReConnectClient>d__64.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   at HaniYar.TLSharpMethods.<SendForwardMessageLast>d__74.MoveNext().

[...]

Issue #668, opened 2017-12-23
msg_id too low...


deadmann:
Starting 2 Hour Ago, i'm receiving an exception. 
_msg_seqno too low (the server has already received a message with a lower msg_id but with either a higher or an equal and odd seqno)_

   at TLSharp.Core.Network.MtProtoSender.HandleBadMsgNotification(UInt64 messageId, Int32 sequence, BinaryReader messageReader)
   at TLSharp.Core.Network.MtProtoSender.processMessage(UInt64 messageId, Int32 sequence, BinaryReader messageReader, TLMethod request)
   at TLSharp.Core.Network.MtProtoSender.<Receive>d__9.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   at TLSharp.Core.TelegramClient.<ConnectAsync>d__9.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   at TelegramAdvertiser.Helper.TelegramHelper.<NewClientAsync>d__12.MoveNext() in C:\Users\Hassan\Documents\Visual Studio 2015\Projects\Advertiser\TelegramAdvertiser\Helper\TelegramHelper.cs:line 161
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   at TelegramAdvertiser.Helper.TelegramHelper.<FindUserByPhoneNumberAsync>d__8.MoveNext() in C:\Users\Hassan\Documents\Visual Studio 2015\Projects\Advertiser\TelegramAdvertiser\Helper\TelegramHelper.cs:line 64

[...]

PR #910, opened 2020-02-22
add handler for exception with code 32

n3nikita:
Add handler for exception with code 32. #668 and #777 issues.

### DIFF
--- a/TLSharp.Core/Network/MtProtoSender.cs
+++ b/TLSharp.Core/Network/MtProtoSender.cs
@@ -205,7 +205,7 @@ namespace TLSharp.Core.Network
                     return HandleBadServerSalt(messageId, sequence, messageReader, request, token);
                 case 0xa7eff811: // bad_msg_notification
                                  //logger.debug("MSG bad_msg_notification");
-                    return HandleBadMsgNotification(messageId, sequence, messageReader);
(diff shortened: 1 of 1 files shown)


knocte:
What does this achieve? It just says "add handler", does this mean it is the fix for #668 and #777?

n3nikita:
Yes. I've just PR the code from issue #668 which should fix error with code 32.

knocte:
This is very weird code.

@CheshireCaat can you review?

CheshireCaat:
I think there is a more correct way to get rid of this error. How can I repeat it to make sure that this solution is right?

CheshireCaat:
Okay, I looked at issues related to this problem, I think we need a more global solution for working with parallel requests as well. 
Maybe it will be better to add something like SemaphoreSlim to the `SendRequestAsync` or `RequestWithDcMigration`?

CheshireCaat:
Now the code from the example below will throw errors
```csharp
await Task.WhenAll(Enumerable.Range(1, 5).Select(arg => Client.CheckUsernameAsync("CheshireCaat")))
```
I tested it also with modified `RequestWithDcMigration` with `SemaphoreSlim` - it works fine.
Can you check your error by adding Semaphore? If this helps, I think it's better to do it that way.

CheshireCaat:
@n3nikita ping

PR #919, opened 2020-04-01
[WIP] Asynchronous Network  

aarani:
I created this as a PoC it works but it's not great I want to propose a new branch for everyone to work on this as it's extremely necessary for having a functional telegram api 
PS: I use NetCoreServer library and it needs a .NET Framework 4.6 Update


aarani:
@knocte Please check this commit and tell me your opinion

TLSharp.Core/Network/TcpTransport.cs
```diff
@@ -10,103 +12,64 @@ namespace TLSharp.Core.Network
...
+        private readonly BlockingCollection<TcpMessage> ReceievedMessage = new BlockingCollection<TcpMessage>();
+        //private readonly NetworkStream stream;
         private int sendCounter = 0;
 
         public TcpTransport(string address, int port, TcpClientConnectionHandler handler = null)
         {
-            if (handler == null)
-            {
-                var ipAddress = IPAddress.Parse(address);
-                var endpoint = new IPEndPoint(ipAddress, port);
```
aarani, line 22: As far as i know this library does not handle handlers and proxy so we should work on an implementation later.

TLSharp.Core/Network/TcpTransport.cs
```diff
@@ -10,103 +12,64 @@ namespace TLSharp.Core.Network
...
-
-            byte[] rv = new byte[packetLengthBytes.Length + seqBytes.Length + body.Length];
-
-            Buffer.BlockCopy(packetLengthBytes, 0, rv, 0, packetLengthBytes.Length);
-            Buffer.BlockCopy(seqBytes, 0, rv, packetLengthBytes.Length, seqBytes.Length);
-            Buffer.BlockCopy(body, 0, rv, packetLengthBytes.Length + seqBytes.Length, body.Length);
-            var crc32 = new Ionic.Crc.CRC32();
-            crc32.SlurpBlock(rv, 0, rv.Length);
-            var validChecksum = crc32.Crc32Result;
+            return ReceievedMessage.Take();
```
aarani, line 52: Old Code Compatibility Only We need to get rid of this

knocte:
@CheshireCaat can you test this?

TLSharp.Core/Network/TcpTransport.cs
```diff
@@ -10,103 +12,64 @@ namespace TLSharp.Core.Network
...
+        }
 
-                tcpClient = new TcpClient(ipAddress.AddressFamily);
-                tcpClient.Connect(endpoint);
-            }
-            else
-                tcpClient = handler(address, port);
+        public void ReceiveMessage(TcpMessage message)
+        {
+            ReceievedMessage.Add(message);
```
aarani, line 34: Old Code Compatibility. It's here to make sync receive function work.
In the final version we should read the message deserialize it find the type and call the handler for that type.
for example if we get a TLMessage we should search a dictionary<type,delegate> to find the handler function.

knocte:
@aarani : to begin with, can you separate this into 2 PRs? another one which just upgrades the .NETFramework version

aarani:
> @CheshireCaat can you test this?

I should say that i used a console application to test it because i couldn't update the test projects to .net framework 4.6

aarani:
> @aarani : to begin with, can you separate this into 2 PRs? another one which just upgrades the .NETFramework version

Sure, I try doing it but test projects are pain to update

knocte:
> because i couldn't update the test projects to .net framework 4.6

why you couldn't upgrade them?

knocte:
> Sure, I try doing it but test projects are pain to update

why a pain? please share more information with us, we're also developers, not managers :)

aarani:
> > because i couldn't update the test projects to .net framework 4.6
> 
> why you couldn't upgrade them?

Visual studio gave me a weird error when I tried doing it so I did not bother to figure out what's wrong. I try again.

Capture

aarani:
> > Sure, I try doing it but test projects are pain to update
> 
> why a pain? please share more information with us, we're also developers, not managers :)

Sure, but I wanted to take a look at it myself first and then bother you guys with it 😄

knocte:
> Visual studio gave me a weird error when I tried doing it so I did not bother to figure out what's wrong. I try again.

I think this happened me once, when I wrote "4.6.0" instead of "4.6" (the latter is the correct version number). Maybe this was the mistake?

If not, try simply restarting Visual Studio like this guy says here: https://stackoverflow.com/a/56597422/544947

aarani:
> > Visual studio gave me a weird error when I tried doing it so I did not bother to figure out what's wrong. I try again.
> 
> I think this happened to me once, when I wrote "4.6.0" instead of "4.6" (the latter is the correct version number). Maybe this was the mistake?
> 
> If not, try simply restarting Visual Studio like this guy says here: https://stackoverflow.com/a/56597422/544947

wow restarting worked 😆 I will try running the tests too.

aarani:
> @aarani : to begin with, can you separate this into 2 PRs? another one which just upgrades the .NETFramework version

#920

aarani:
Sorry, I accidentally pushed my api_id and api_hash and had to force push it this pull_request is toast

PR #920, opened 2020-04-01
Update to .NET Framework 4.6

aarani:


knocte:
@aarani cool stuff! btw, while you're at it, I was thinking... before we merge this change, we should make sure that it doesn't break the build in nonWindows OSs. For this, I was pondering migrating from AppVeyor to GithubActionsCI because the latter provides free-tier for macOS&Linux too, have you ever tried it?

aarani:
> @aarani cool stuff! btw, while you're at it, I was thinking... before we merge this change, we should make sure that it doesn't break the build in nonWindows OSs. For this, I was pondering migrating from AppVeyor to GithubActionsCI because the latter provides free-tier for macOS&Linux too, have you ever tried it?

Not really. Do you mean .net core or mono ?

aarani:
Because It's actually quite some times I'm thinking about migrating this code to .net core. I already have alternatives too all libraries that we use.

knocte:
> Do you mean .net core or mono ?

Actually both macOS and Linux VMs that GithubActionsCI use come with both mono and .NETCore preinstalled.

>  I'm thinking about migrating this code to .net core.

TLSharp is a library and, as such, it doesn't need to be migrated to .NETCore. It can be migrated to .NETStandard so that it can be used both from .NETCore *and* .NETFramework.

However, I was not talking about .NETStandard yet, I think we can do that conversion later. I think getting CI to work in a cross-platform way first, is more important.

aarani:
> > Do you mean .net core or mono ?
> 
> Actually both macOS and Linux VMs that GithubActionsCI use come with both mono and .NETCore preinstalled.
> 
> > I'm thinking about migrating this code to .net core.
> 
> TLSharp is a library and, as such, it doesn't need to be migrated to .NETCore. But It can be migrated to .NETStandard so that it can be used both from .NETCore **and** .NETFramework.
> 
> However, I was not talking about .NETStandard yet, I think we can do that conversion later. I think getting CI to work in a cross-platform way first, is more important.

Is Mono Build currently broken or you think updating to 4.6 will break it ?

knocte:
> Is Mono Build currently broken

I don't think so. The last time I tried manually, it worked.

> or you think updating to 4.6 will break it ?

I don't think it would break it, but having CI before merging this PR would make us be sure about this.

aarani:
> > Is Mono Build currently broken
> 
> I don't think so. The last time I tried manually, it worked.
> 
> > or you think updating to 4.6 will break it ?
> 
> I don't think it would break it, but having CI before merging this PR would make us be sure about this.

Ok. let me create a fork and try adding CI and check it.

aarani:
> > Is Mono Build currently broken
> 
> I don't think so. The last time I tried manually, it worked.
> 
> > or you think updating to 4.6 will break it ?
> 
> I don't think it would break it, but having CI before merging this PR would make us be sure about this.

Ok I will do some searching for how to write a workflow for mono but I never done it before.

aarani:
@knocte Check #921 but looks like github actions support windows too

aarani:
CI Status for this specific pull request for example : https://github.com/aarani/TLSharp/pull/2

knocte:
> but looks like github actions support windows too

Of course, I said GithubActions CI is crossplatform (so not Windows exclusively like AppVeyor, but Linux&macOS too).

aarani:
@knocte what about this?

knocte:
Can you rebase the PR?

aarani:
I don't think it needs that.

aarani:
I don't have access to this pull request's branch unless I update my repo and create another pull request. Rebasing this should be possible just using github ui.

knocte:
Yes, otherwise we cannot see the CI status of the PR.

aarani:
Ok

aarani:
See #922

PR #921, opened 2020-04-01
Github Action Workflow

aarani:


aarani:
@knocte Do you have permission to delete this pull request?

aarani:
@knocte https://github.com/aarani/TLSharp/runs/551719993

knocte:
Good job! Let's add macOS too (in macOS you can use msbuild, no need to use xbuild).

aarani:
> Good job! Let's add macOS too (in macOS you can use msbuild, no need to use xbuild).

macOS msbuild doesn't support .NET Framework afaik

knocte:
> macOS msbuild doesn't support .NET Framework afaik

But I mean Mono. And macOS+mono already includes msbuild (which is now opensource).

knocte:
> macOS msbuild doesn't support .NET Framework afaik

I meant that it does support it, in fact msbuild is actually the .NET framework tool (the .NET Core tool to build is "dotnet build" ;) ).

.github/workflows/main.yml
```diff
@@ -0,0 +1,61 @@
...
+
+name: CI Build
+
+# Controls when the action will run. Triggers the workflow on push or pull request
+# events but only for the master branch
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
```
knocte, line 11: it's better to not specify master here, so that CI works in any branch; so I would replace this with: `on: [push, pull_request]`

knocte:
Great! this is almost ready, I'll now just point out some other nitpicks

.github/workflows/main.yml
```diff
@@ -0,0 +1,79 @@
...
+    - uses: actions/checkout@v2
+    - name: wget
+      uses: wei/wget@v1
+      with:
+        args: -O nuget.exe https://dist.nuget.org/win-x86-commandline/latest/nuget.exe
+    
+    - name: Install Dependencies
+      run: |
+        mono nuget.exe install
+        mono nuget.exe restore
```
knocte, line 32: I think that only the restore is needed, can you try removing the `mono nuget.exe install` line?

.github/workflows/main.yml
```diff
@@ -0,0 +1,79 @@
...
+    runs-on: ubuntu-latest
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+    # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+    - uses: actions/checkout@v2
+    - name: wget
+      uses: wei/wget@v1
+      with:
+        args: -O nuget.exe https://dist.nuget.org/win-x86-commandline/latest/nuget.exe
```
knocte, line 23: I think it's simpler if you just install wget with `sudo apt install -y wget` and just use it, instead of depending on an external action

aarani: wget is installed by default but it gives HTTP 404 Error when you try to download anything

.github/workflows/main.yml
```diff
@@ -0,0 +1,69 @@
...
+    - uses: actions/checkout@v2
+    - name: Download latest NuGet (cross-platform)
+      uses: olegtarasov/download-nuget@v1
+
+    - name: Install Dependencies
+      run: mono $NUGET_EXE restore
+    
+    # Runs a set of commands using the runners shell
+    - name: Build Project
+      run: xbuild TLSharp.Core/TLSharp.Core.csproj
```
knocte, line 47: in macOS let's use msbuild (cause xbuild is deprecated)

.github/workflows/main.yml
knocte: how can it give 404? I use it all the time myself, maybe switch to curl?

.github/workflows/main.yml
```diff
@@ -0,0 +1,70 @@
...
+    - name: Download latest NuGet (cross-platform)
+      uses: olegtarasov/download-nuget@v1
+
+    - name: Install Dependencies
+      run: |
+         mono $NUGET_EXE restore
+    
+    # Runs a set of commands using the runners shell
+    - name: Build Project
+      run: MSBuild TLSharp.Core/TLSharp.Core.csproj
```
knocte, line 47: msbuild should be able to build the .sln file? (not sure about xbuild, but let's try too)

aarani: xbuild can build .sln but it fails when trying to build Test.VS

.github/workflows/main.yml
aarani: Take a look:
https://github.com/aarani/TLSharp/runs/551697235?check_suite_focus=true
https://github.com/aarani/TLSharp/runs/551693010?check_suite_focus=true

.github/workflows/main.yml
knocte: then let's use xbuild TLSharp.Core.csproj in Linux and msbuild TLSharp.sln in the others

aarani: https://github.com/aarani/TLSharp/runs/551704535?check_suite_focus=true

aarani: macos fails to compile VS Unit Tests too. It's because of the mono not xbuild.
https://github.com/aarani/TLSharp/runs/554069446?check_suite_focus=true

.github/workflows/main.yml
knocte: did you add a comma at the end of the URL by mistake?

.github/workflows/main.yml
knocte: ok then let's just compile the .sln on Windows

.github/workflows/main.yml
aarani: Yep you're right i thought it's yaml syntax at first but it wasn't

knocte: great, let's download nuget.exe manually then

.github/workflows/main.yml
aarani: Done

.github/workflows/main.yml
aarani: Done

.github/workflows/main.yml
```diff
@@ -0,0 +1,48 @@
...
+    - name: Build Project
+      run: xbuild TLSharp.Core/TLSharp.Core.csproj
+      
+  macos:
+    runs-on: macos-latest
+    steps:
+    - uses: actions/checkout@v2
+      
+    - name: Download latest NuGet
+      uses: olegtarasov/download-nuget@v1
```
knocte, line 25: same here? maybe curl or wget is already installed in macOS?

aarani: Done.

.github/workflows/main.yml
```diff
@@ -0,0 +1,48 @@
...
+    - name: Build Project
+      run: MSBuild TLSharp.Core/TLSharp.Core.csproj
+
+  windows:
+    runs-on: windows-latest
+    steps:
+    - uses: actions/checkout@v2
+    
+    - name: Setup Nuget.exe
+      uses: warrenbuckley/Setup-Nuget@v1
```
knocte, line 39: ok, now I believe Windows doesn't have wget or curl by default, but you can use powershell to download a file, see: https://stackoverflow.com/a/51225744/544947

aarani: https://github.com/aarani/TLSharp/runs/554107473?check_suite_focus=true

knocte:
beautiful! is there a way to have variables in the workflow to not repeat the nuget URL 3 times?

.github/workflows/main.yml
aarani: i think it's saving the file somewhere else

aarani: it's failing

aarani:
> beautiful! is there a way to have variables in the workflow to not repeat the nuget URL 3 times?

Done but downloading using powershell still not working

knocte:
I've had problems in the past with regressions in the newest nuget versions (and suddenly the CI breaking without changing anything can be scary&annoying) so let's switch to a specific version: I recommend 5.4.0

aarani:
Fixed

.github/workflows/main.yml
aarani: Fixed

aarani: Fixed

aarani:
> I've had problems in the past with regressions in the newest nuget versions (and suddenly the CI breaking without changing anything can be scary&annoying) so let's switch to a specific version: I recommend 5.4.0

Done

aarani:
When using environment variables to feed URL to Invoke-WebRequest It fails

.github/workflows/main.yml
```diff
@@ -0,0 +1,52 @@
+name: CI Build
```
knocte, line 1: one last thing, I think it's better to have the name of the yml file match the name of the workflow, so let's rename this from "CI Build" to just "CI", and let's rename the file from "main.yml" to "CI.yml", this way it's easier to add badges to the readme file

aarani:
Downloading NuGET on windows when using environment variables fails. any idea why ?

knocte:
> Downloading NuGET on windows when using environment variables fails. any idea why ?

not sure, I'll investigate

.github/workflows/CI.yml
```diff
@@ -0,0 +1,50 @@
...
+    - name: Build Project
+      run: MSBuild TLSharp.Core/TLSharp.Core.csproj
+
+  windows:
+    runs-on: windows-latest
+    steps:
+    - uses: actions/checkout@v2
+    
+    - name: Download Nuget.exe
+      run: Invoke-WebRequest -OutFile Nuget.exe https://dist.nuget.org/win-x86-commandline/v5.4.0/nuget.exe
```
knocte, line 41: maybe the syntax is `$Env:NUGET_URL` ? according to https://www.computerperformance.co.uk/powershell/environmental-variables/

aarani:
Yes that worked. 
Final Log: https://github.com/aarani/TLSharp/actions/runs/68830282

knocte:
Now let's add the badge for the build status, in the readme file, and I merge :)

In markdown format should be: [CI build status badge]

knocte:
(And remove the WIP from the title?)

aarani:
@knocte Can you please remove the appvoyer check?

PR #924, opened 2020-04-03
Added an IPv4/IPv6 toggle to TelegramClient constructor

solarin:
that allows choosing whether to accept ipv4 or ipv6 or both ip address versions passed by the Telegram server.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1,0 +1,50 @@
+name: CI
+on: [push, pull_request]
+
+env:
+  NUGET_URL: https://dist.nuget.org/win-x86-commandline/v5.4.0/nuget.exe
+jobs:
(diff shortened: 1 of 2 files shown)


knocte:
Can you revert the change to the .sln file? It's what's causing the build to fail.

aarani:
You added a project to the solution that you added to .gitignore.
First of all please revert the changes to gitignore and second exclude the project from Solution file.

solarin:
sure, let me do it. sorry guys, i haven't used github frequently

solarin:
by the way, if you guys agree I would rename the enum to DataCenterIPvPrefs or DataCenterIPVersions

solarin:
moved to 925 https://github.com/sochix/TLSharp/pull/925/

PR #925, opened 2020-04-03
Added an IPv4/IPv6 toggle to TelegramClient constructor 

solarin:


knocte:
@solarin now you forgot to include the changes to the .csproj file :)

And please configure your editor to use 4 spaces, not tabs.

solarin:
> @solarin now you forgot to include the changes to the .csproj file :)
> And please configure your editor to use 4 spaces, not tabs.

i'm getting crazy

solarin:
also added some methods that i needed for my project


+ GetFullChat to get the info of a chat 
+ GetAllChats to get all available user's chats
+ MakeAuthBotAsync to authenticate a bot instead of a user

knocte:
> also added some methods that i needed for my project

Please don't include those here, open a different PullRequest for that.

TLSharp.Core/DataCenterIPVersion.cs
```diff
@@ -0,0 +1,30 @@
...
+        /// <summary>
+        /// Takes only IPv6 addresses
+        /// </summary>
+        OnlyIPv6 =   2,
+        /// <summary>
+        /// Connection to IPv4 addresses is preferred to IPv6 addresses
+        /// </summary>
+        PreferIPv4 = 3,
+        /// <summary>
+        /// Connection to IPv6 addresses is preferred to IPv6 addresses
```
knocte, line 26: typo here ^

TLSharp.Core/DataCenterIPVersion.cs
```diff
@@ -0,0 +1,30 @@
+﻿namespace TLSharp.Core
+{
+    /// <summary>
+    /// When the Telegram server responds with a set of addresses to connect to, DataCenterIPVersion indicates a preference 
+    /// for how to choose the IP address to connect to 
+    /// </summary>
+	public enum DataCenterIPVersion
+    {
+        /// <summary>
+        /// Prefers IPv6 addresses if any is passed by Telegram
```
knocte, line 10: bad copy-paste here ^

TLSharp.Core/DataCenterIPVersion.cs
```diff
@@ -0,0 +1,30 @@
+﻿namespace TLSharp.Core
+{
+    /// <summary>
+    /// When the Telegram server responds with a set of addresses to connect to, DataCenterIPVersion indicates a preference 
+    /// for how to choose the IP address to connect to 
+    /// </summary>
+	public enum DataCenterIPVersion
```
knocte, line 7: this line still has a tab instead of 4 spaces

TLSharp.Core/TelegramClient.cs
```diff
@@ -51,9 +64,10 @@ public Session Session
             this.apiHash = apiHash;
             this.apiId = apiId;
             this.handler = handler;
+            this.dcIpVersion = dcIpVersion;
 
             session = Session.TryLoadOrCreateNew(store, sessionUserId);
-            transport = new TcpTransport (session.DataCenter.Address, session.DataCenter.Port, this.handler);
+            transport = new TcpTransport(session.DataCenter.Address, session.DataCenter.Port, this.handler);
```
knocte, line 70: no need to change this line above ^

TLSharp.Core/TelegramClient.cs
```diff
@@ -123,15 +151,15 @@ private async Task RequestWithDcMigration(TLMethod request, CancellationToken to
                 throw new InvalidOperationException("Not connected!");
 
             var completed = false;
-            while(!completed)
+            while (!completed)
```
knocte, line 154: same here ^

TLSharp.Core/TelegramClient.cs
```diff
@@ -123,15 +151,15 @@ private async Task RequestWithDcMigration(TLMethod request, CancellationToken to
...
             {
                 try
                 {
                     await sender.Send(request, token).ConfigureAwait(false);
                     await sender.Receive(request, token).ConfigureAwait(false);
                     completed = true;
                 }
-                catch(DataCenterMigrationException e)
+                catch (DataCenterMigrationException e)
```
knocte, line 162: etc...

TLSharp.Core/TelegramClient.cs
solarin: when i swapped the editor to replace tabs with the spaces, i had to autoformat and this was auto formatted as well.

TLSharp.Core/TelegramClient.cs
solarin: same as above

TLSharp.Core/TelegramClient.cs
solarin: same as above

TLSharp.Core/DataCenterIPVersion.cs
solarin: fixed

solarin:
> also added some methods that i needed for my project
> 
> Please don't include those here, open a different PullRequest for that.

ok i deleted these and created a new PR

TLSharp.Core/DataCenterIPVersion.cs
knocte: this one is not fixed

TLSharp.Core/TelegramClient.cs
knocte: well then revert the autoformat, you're breaking the git-blame for the lines that don't need changing

TLSharp.Core/DataCenterIPVersion.cs
solarin: strange, i changed it O.o

solarin:
i think i have changed all the spaces back to the originals

TLSharp.Core/DataCenterIPVersion.cs
knocte: yes, this one was fixed, but I still see the text “Prefers IPv6 addresses if any is passed by Telegram” describing “Default”

knocte:
> i think i have changed all the spaces back to the originals

You missed many, sorry, click here and you will see them all: https://github.com/sochix/TLSharp/pull/925/files

TLSharp.Core/DataCenterIPVersion.cs
solarin: it's what actually does

knocte: No, the text “Prefers IPv6 addresses if any is passed by Telegram” should describe "PreferIPv6", not "Default".

knocte: remember Default doesn't do any filtering, just chooses the first datacenter's IP given by telegram servers

TLSharp.Core/DataCenterIPVersion.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// <summary>
+    /// When the Telegram server responds with a set of addresses to connect to, DataCenterIPVersion indicates a preference 
+    /// for how to choose the IP address to connect to 
+    /// </summary>
+    public enum DataCenterIPVersion
+    {
+        /// <summary>
+        /// Prefers IPv6 addresses if any is passed by Telegram
+        /// </summary>
+        Default = 0,
```
knocte, line 12: why do we need this enum be int-based? there's no need

solarin: there is no need.

knocte: cool let's remove the assignments to integers then

TLSharp.Core/DataCenterIPVersion.cs
solarin: it doesn't do any filtering, but it's a side effect of the algorithm to pick an ipv6 if any is passed just because it will be at the end of the list.

anyways, because the default behavious might change in the future, we can remove this comment.

knocte: we can fix the side-effect, don't do any OrderBy if .Default is chosen please.

knocte:
Thanks for your patience. I will merge this tonite or tomorrow.

solarin:
i propose to rename DataCenterIPVersion to DataCenterIPVersions
and also move this file under namespace TLSharp.Core.Types

because we need other enums and other types that we need logically, although they have a different name or type at protocol-level

One example is ParticipantTypes (Recents, 
        Restricted,
        Admins,
        Bots,
        Search,
        Contacts,
        Kicked,
        Banned)

which at protcol level exists in different ways.

SALEH-SOFTWARE:
var client = new TelegramClient(apiId, apiHash);
   await client.ConnectAsync(); 

this code not have dependency !! How Can Now Connect To Telegram? And Please Put Poll Method For Vote Channel

knocte:
> this code not have dependency !

what dependency? what are you talking about?

PR #926, opened 2020-04-03
GetFullChat + GetAllChats + MakeAuthBotAsync 

solarin:
+ GetFullChat to get the info of a chat
+ GetAllChats to get all available user's chats
+ MakeAuthBotAsync to authenticate a bot instead of a user

### DIFF
--- a/TLSharp.Core/TelegramClient.cs
+++ b/TLSharp.Core/TelegramClient.cs
@@ -25,458 +25,458 @@ using TLAuthorization = TeleSharp.TL.Auth.TLAuthorization;
 namespace TLSharp.Core
 {
 
-	public class TelegramClient : IDisposable
-	{
-		private MtProtoSender sender;
-		private TcpTransport transport;
-		private string apiHash = String.Empty;
-		private int apiId = 0;
-		private Session session;
-		private List<TLDcOption> dcOptions;
-		private TcpClientConnectionHandler handler;
-		private DataCenterIPVersion dcIpVersion;
-
-		public Session Session
-		{
-			get { return session; }
-		}
-
-		/// <summary>
-		/// Creates a new TelegramClient
-		/// </summary>
-		/// <param name="apiId">The API ID provided by Telegram. Get one at https://my.telegram.org </param>
-		/// <param name="apiHash">The API Hash provided by Telegram. Get one at https://my.telegram.org </param>
-		/// <param name="store">An ISessionStore object that will handle the session</param>
-		/// <param name="sessionUserId">The name of the session that tracks login info about this TelegramClient connection</param>
-		/// <param name="handler">A delegate to invoke when a connection is needed and that will return a TcpClient that will be used to connect</param>
-		/// <param name="dcIpVersion">Indicates the preferred IpAddress version to use to connect to a Telegram server</param>
-		public TelegramClient(int apiId, string apiHash,
-			ISessionStore store = null, string sessionUserId = "session", TcpClientConnectionHandler handler = null, DataCenterIPVersion dcIpVersion = DataCenterIPVersion.Default)
-		{
-			if (apiId == default(int))
-				throw new MissingApiConfigurationException("API_ID");
-			if (string.IsNullOrEmpty(apiHash))
-				throw new MissingApiConfigurationException("API_HASH");
-
-			if (store == null)
-				store = new FileSessionStore();
-
-			this.apiHash = apiHash;
-			this.apiId = apiId;
-			this.handler = handler;
-			this.dcIpVersion = dcIpVersion;
-
-			session = Session.TryLoadOrCreateNew(store, sessionUserId);
-			transport = new TcpTransport(session.DataCenter.Address, session.DataCenter.Port, this.handler);
-		}
-
-		public async Task ConnectAsync(bool reconnect = false, CancellationToken token = default(CancellationToken))
-		{
-			token.ThrowIfCancellationRequested();
-
-			if (session.AuthKey == null || reconnect)
-			{
-				var result = await Authenticator.DoAuthentication(transport, token).ConfigureAwait(false);
-				session.AuthKey = result.AuthKey;
-				session.TimeOffset = result.TimeOffset;
-			}
-
-			sender = new MtProtoSender(transport, session);
-
-			//set-up layer
-			var config = new TLRequestGetConfig();
-			var request = new TLRequestInitConnection()
-			{
-				ApiId = apiId,
-				AppVersion = "1.0.0",
-				DeviceModel = "PC",
-				LangCode = "en",
-				Query = config,
-				SystemVersion = "Win 10.0"
-			};
-			var invokewithLayer = new TLRequestInvokeWithLayer() { Layer = 66, Query = request };
-			await sender.Send(invokewithLayer, token).ConfigureAwait(false);
-			await sender.Receive(invokewithLayer, token).ConfigureAwait(false);
-
-			dcOptions = ((TLConfig)invokewithLayer.Response).DcOptions.ToList();
-		}
-
-		private async Task ReconnectToDcAsync(int dcId, CancellationToken token = default(CancellationToken))
-		{
-			token.ThrowIfCancellationRequested();
-
-			if (dcOptions == null || !dcOptions.Any())
-				throw new InvalidOperationException($"Can't reconnect. Establish initial connection first.");
-
-			TLExportedAuthorization exported = null;
-			if (session.TLUser != null)
-			{
-				TLRequestExportAuthorization exportAuthorization = new TLRequestExportAuthorization() { DcId = dcId };
-				exported = await SendRequestAsync<TLExportedAuthorization>(exportAuthorization, token).ConfigureAwait(false);
-			}
-
-			var dcs = dcOptions.Where(d => d.Id == dcId
-											&& (
-												(dcIpVersion == DataCenterIPVersion.Default) // any
-												|| (d.Ipv6 && dcIpVersion == DataCenterIPVersion.OnlyIPv6) // selects only ipv6 addresses 	
-												|| (!d.Ipv6 && dcIpVersion == DataCenterIPVersion.OnlyIPv4) // selects only ipv4 addresses
-												|| dcIpVersion == DataCenterIPVersion.PreferIPv4 // we can take both types of address
-												|| dcIpVersion == DataCenterIPVersion.PreferIPv6 // we can take both types of address
-												)
-											).OrderBy(d => d.Ipv6);
-
-			if (dcs.Count() == 0)
-				throw new Exception($"Telegram server didn't provide us with any IPAddress that matches your preferences. If you chose OnlyIPvX, try switch to PreferIPvX instead.");
+    public class TelegramClient : IDisposable
+    {
+        private MtProtoSender sender;
+        private TcpTransport transport;
+        private string apiHash = String.Empty;
+        private int apiId = 0;
+        private Session session;
+        private List<TLDcOption> dcOptions;
+        private TcpClientConnectionHandler handler;
+        private DataCenterIPVersion dcIpVersion;
+
+        public Session Session
+        {
+            get { return session; }
+        }
+
+        /// <summary>
+        /// Creates a new TelegramClient
+        /// </summary>
+        /// <param name="apiId">The API ID provided by Telegram. Get one at https://my.telegram.org </param>
+        /// <param name="apiHash">The API Hash provided by Telegram. Get one at https://my.telegram.org </param>
+        /// <param name="store">An ISessionStore object that will handle the session</param>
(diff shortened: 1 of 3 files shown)


knocte:
You need to remove the DataCenterIPVersion stuff from here, otherwise the PRs are not separate (if I merge the other one, there will be conflicts here).

solarin:
> You need to remove the DataCenterIPVersion stuff from here, otherwise the PRs are not separate (if I merge the other one, there will be conflicts here).

ok i think it's ok now

knocte:
@aarani can you review this PR?

TLSharp.Core/TelegramClient.cs
```diff
@@ -406,6 +414,221 @@ public async Task<TLAbsMessages> GetHistoryAsync(TLAbsInputPeer peer, int offset
...
+                case ParticipantTypes.Recents:
+                    filter = new TLChannelParticipantsRecent();
+                    break;
+
+                case ParticipantTypes.Banned:
+                case ParticipantTypes.Restricted:
+                case ParticipantTypes.Contacts:
+                case ParticipantTypes.Search:
+                default:
+                    throw new NotImplementedException($"{partType} not implemented yet");
```
aarani, line 570: If it's not gonna be implemented please remove them from your enum as the only purpose of enum is for this filter

TLSharp.Core/TelegramClient.cs
```diff
@@ -406,6 +414,221 @@ public async Task<TLAbsMessages> GetHistoryAsync(TLAbsInputPeer peer, int offset
...
+        /// </summary>
+        /// <param name="channelId">The id of the channel whose participants are requested</param>
+        /// <param name="accessHash">The access hash of the channel whose participants are requested</param>
+        /// <param name="stIdx">The index to start fetching from. -1 will automatically fetch all the results</param>
+        /// <param name="pageSize">How many results to be fetch on each iteration. 
+        /// Values smaller than 0 are ignored. If stIdx is set, a number of results smaller than pageSize might be returned by Telegram.</param>
+        /// <param name="partType">The type of the participants to get. Choose Recents not to filter</param>
+        /// <param name="token"></param>
+        /// <returns></returns>
+        public async Task<TLChannelParticipants> GetParticipants(int channelId, long accessHash, int stIdx = -1, int pageSize = -1, ParticipantTypes partType = ParticipantTypes.Recents, CancellationToken token = default(CancellationToken))
```
aarani, line 544: change partType to something like filter so it's possible to understand the purpose without reading function documention

TLSharp.Core/Types/ParticipantTypes.cs
```diff
@@ -0,0 +1,16 @@
+﻿namespace TLSharp.Core.Types
+{
+    public enum ParticipantTypes
```
aarani, line 3: Change this to something like ParticipantTypesFilter too maybe?

solarin: ParticipantTypeFilters

TLSharp.Core/TelegramClient.cs
solarin: ok

TLSharp.Core/TelegramClient.cs
solarin: it's gonna be implemented as soon as the new Classes are added to the TL framework

solarin: should i drop them anyways or leave them?

TLSharp.Core/TelegramClient.cs
```diff
@@ -406,6 +414,221 @@ public async Task<TLAbsMessages> GetHistoryAsync(TLAbsInputPeer peer, int offset
...
+            return ret;
+        }
+        /// <summary>
+        /// Invites the passed users to the specified channel
+        /// </summary>
+        /// <param name="channelId">The id of the channel to invite the users to</param>
+        /// <param name="accessHash">The access hash of the channel to invite the users to</param>
+        /// <param name="users"></param>
+        /// <returns></returns>
+        public async Task<TLUpdates> InviteToChannel(int channelId, long accessHash, int[] users, CancellationToken token = default(CancellationToken))
```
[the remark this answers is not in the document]
solarin: I create an overload as for the other methods, ok?

TLSharp.Core/TelegramClient.cs
```diff
@@ -406,6 +414,221 @@ public async Task<TLAbsMessages> GetHistoryAsync(TLAbsInputPeer peer, int offset
...
+            do
+            {
+                var req = new TLRequestGetParticipants()
+                {
+                    Channel = new TLInputChannel()
+                    {
+                        ChannelId = channelId,
+                        AccessHash = accessHash
+                    },
+                    Filter = new TLChannelParticipantsRecent(),
```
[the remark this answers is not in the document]
solarin: right, i didn't replace it with the local variable. this was a test.

TLSharp.Core/TelegramClient.cs
```diff
@@ -17,13 +18,20 @@
 using TLSharp.Core.MTProto.Crypto;
 using TLSharp.Core.Network;
 using TLSharp.Core.Network.Exceptions;
+using TLSharp.Core.Types;
 using TLSharp.Core.Utils;
 using TLAuthorization = TeleSharp.TL.Auth.TLAuthorization;
+using TLRequestUpdateUsername = TeleSharp.TL.Account.TLRequestUpdateUsername;
```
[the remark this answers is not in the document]
solarin: actually TLRequestUpdateUsername  is ambigous with as it's already present in TL.Channels. i can double check and report exactly where.

TLSharp.Core/TelegramClient.cs
```diff
@@ -607,6 +607,16 @@ public async Task<TLChannelParticipants> GetParticipants(int channelId, long acc
         /// <summary>
         /// Invites the passed users to the specified channel
         /// </summary>
+        /// <param name="channel">The descriptor of the channel to invite the users to</param>
+        /// <param name="users"></param>
+        /// <returns></returns>       
+        public async Task<TLUpdates> InviteToChannel(TLChannel channel, int[] users, CancellationToken token = default(CancellationToken))
+        {
+            return await InviteToChannel(channel.Id, (long)channel.AccessHash, users, token);
+        }
```
knocte, line 616: please add an EOL here, to separate the 2 methods properly

TLSharp.Core/TelegramClient.cs
```diff
@@ -342,44 +350,44 @@ public async Task<TLAbsDialogs> GetUserDialogsAsync(int offsetDate = 0, int offs
...
-                {
-                    Location = location,
-                    Limit = filePartSize,
-                    Offset = offset
-                }, token)
+            {
+                Location = location,
+                Limit = filePartSize,
+                Offset = offset
+            }, token)
```
knocte, line 390: you formatted the whole file again by mistake, can you create a separate PR just to format the file please?

solarin: see the next commit

TLSharp.Core/TelegramClient.cs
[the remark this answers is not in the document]
aarani: sure nice

TLSharp.Core/TelegramClient.cs
aarani: are they present in the new layers?

TLSharp.Core/Types/ParticipantTypes.cs
aarani: ok

TLSharp.Core/TelegramClient.cs
solarin: yes, see here: https://core.telegram.org/type/ChannelParticipantsFilter

TLSharp.Core/TelegramClient.cs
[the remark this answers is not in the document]
solarin: 'TLRequestUpdateUsername' is an ambiguous reference between 'TeleSharp.TL.Account.TLRequestUpdateUsername' and 'TeleSharp.TL.Channels.TLRequestUpdateUsername'	 

 'TLAuthorization' is an ambiguous reference between 'TeleSharp.TL.Auth.TLAuthorization' and 'TeleSharp.TL.TLAuthorization'	TLSharp.Core

TLSharp.Core/TelegramClient.cs
```diff
@@ -406,6 +414,241 @@ public async Task<TLAbsMessages> GetHistoryAsync(TLAbsInputPeer peer, int offset
...
+                        AccessHash = accessHash
+                    },
+                    Filter = filter,
+                    Offset = found,
+                    Limit = pageSize
+                };
+                var fchat = await SendRequestAsync<TLChannelParticipants>(req, token).ConfigureAwait(false);
+                total = fchat.Count;
+                found += fchat.Participants.Count;
+                results.Add(fchat);
```
solarin, line 606: i must remove this results. it is useless.

TLSharp.Core/TelegramClient.cs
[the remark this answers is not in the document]
knocte: that's fair enough

TLSharp.Core/TelegramClient.cs
knocte: I vote, then, for only merging this PR after we have updated the layer

TLSharp.Core/TelegramClient.cs
```diff
@@ -406,6 +414,239 @@ public async Task<TLAbsMessages> GetHistoryAsync(TLAbsInputPeer peer, int offset
...
+        /// Throws an exception if the authenticated user is a bot.
+        /// </summary> 
+        /// <param name="ids">The IDs of the chats to be returned</param>
+        /// <param name="token"></param>
+        /// <returns>The list of chats opened by the authenticated user</returns>
+        public async Task<TLChats> GetAllChats(int[] exceptdIds, CancellationToken token = default(CancellationToken))
+        {
+            var ichats = new TeleSharp.TL.TLVector<int>(); // we can't pass a null argument to the TLRequestGetChats
+            if (exceptdIds != null)
+                Array.ForEach(exceptdIds, x => ichats.Add(x));
```
knocte, line 475: I'm not a fan of `Array.ForEach` because it hardcodes "Array"; isn't there an IEnumerable-esque way of doing this?

solarin:
@knocte as for the Array.ForEach, the only replacement i am aware of is to do a classic for(int i...). i wouldn't use any other list in this case as they would introduce a worse overhead than the Array.ForEach.

knocte:
> as for the Array.ForEach, the only replacement i am aware of is to do a classic for(int i...). i wouldn't use any other list in this case as they would introduce a worse overhead than the Array.ForEach.

Let's use a normal foreach then. More reasoning here: https://docs.microsoft.com/en-gb/archive/blogs/ericlippert/foreach-vs-foreach

TLSharp.Core/TelegramClient.cs
jamesjsc: I used GetParticipants(channelId, accessHash,-1,-1), but GetParticipants always return maximum 10k members. 
How to get all member in a group with over 10k members?

TLSharp.Core/TelegramClient.cs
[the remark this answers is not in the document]
jamesjsc: This function is not working, always return UserId_Invalid exception
You have to assign both UserId and Hash to TLInputUser.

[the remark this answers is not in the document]
knocte: @jamesjsc do you mind porting this PR to TgSharp?

solarin:
this is a limit set by Telegram, there is no such a method to get ALL the
members, in fact this method gets the MOST RECENT members which return 10k
users at most.

TLSharp.Core/TelegramClient.cs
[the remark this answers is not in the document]
jamesjsc: @knocte ok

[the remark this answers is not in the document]
jamesjsc: @knocte What is the different between TLSharp and TgSharp?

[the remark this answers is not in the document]
knocte: TL Layer update, Tg is the future, TL will just receive bug fixes, no new features.

TLSharp.Core/TelegramClient.cs
barrywimlett: var ichats = exceptdIds!=null ? new TeleSharp.TL.TLVector<int>(exceptdIds) : new TeleSharp.TL.TLVector<int>(); 

//    if (exceptdIds != null)
//     Array.ForEach(exceptdIds, x => ichats.Add(x));

PR #927, opened 2020-04-05
IsConnected now includes a check on sender==null.

solarin:
Otherwise callls to this method always return True and if(!TelegramClient.IsConnected) TelegramClient.ConnectAsync() wouldn't have the expected result.


knocte:
hey @salarcode sorry to bother you, but as you implemented this IsConnected property, I thought you might be interested in reviewing this PR? thanks in advance

solarin:
moved to #931

PR #931, opened 2020-04-08
addresses connection issues

solarin:
! IsConnected is restored as it was before
+ TcpTransport has got a new Connect method which properly initializes the underlying tcpclient, this makes it possible to reconnect in case the tcpClient gets disconnected, without restarting the application or recreating a new TelegramClient object.

### DIFF
--- a/TLSharp.Core/Network/TcpTransport.cs
+++ b/TLSharp.Core/Network/TcpTransport.cs
@@ -12,23 +12,36 @@ namespace TLSharp.Core.Network
 
     public class TcpTransport : IDisposable
     {
-        private readonly TcpClient tcpClient;
-        private readonly NetworkStream stream;
+        private TcpClient tcpClient;
+        private NetworkStream stream;
         private int sendCounter = 0;
+        TcpClientConnectionHandler handler;
+        string address;
+        int port;
+        IPAddress ipAddress; 
 
         public TcpTransport(string address, int port, TcpClientConnectionHandler handler = null)
         {
+            this.handler = handler;
+            this.address = address;
+            this.port = port;
+            ipAddress = IPAddress.Parse(address);
+        }
+
+        public async Task Connect()
+        {
             if (handler == null)
             {
-                var ipAddress = IPAddress.Parse(address);
-                var endpoint = new IPEndPoint(ipAddress, port);
-
+                if (tcpClient != null)
+                {
+                    tcpClient.Close();
+                }
                 tcpClient = new TcpClient(ipAddress.AddressFamily);
 
                 try {
-                    tcpClient.Connect (endpoint);
+                    await tcpClient.ConnectAsync(ipAddress, port);
                 } catch (Exception ex) {
-                    throw new Exception ($"Problem when trying to connect to {endpoint}; either there's no internet connection or the IP address version is not compatible (if the latter, consider using DataCenterIPVersion enum)",
+                    throw new Exception ($"Problem when trying to connect to {ipAddress}:{port}; either there's no internet connection or the IP address version is not compatible (if the latter, consider using DataCenterIPVersion enum)",
                                          ex);
                 }
             }
@@ -102,7 +115,7 @@ namespace TLSharp.Core.Network
         {
             get
             {
-                return this.tcpClient.Connected;
+                return this.tcpClient != null && this.tcpClient.Connected;
             }
         }
 
--- a/TLSharp.Core/TelegramClient.cs
+++ b/TLSharp.Core/TelegramClient.cs
@@ -72,6 +72,11 @@ namespace TLSharp.Core
         {
             token.ThrowIfCancellationRequested();
 
+            if (!transport.IsConnected)
+                await transport.Connect();
+            if (!transport.IsConnected)
+                throw new Exception("Connection to Telegram failed");
+
             if (session.AuthKey == null || reconnect)
             {
                 var result = await Authenticator.DoAuthentication(transport, token).ConfigureAwait(false);


TLSharp.Core/Network/TcpTransport.cs
```diff
@@ -12,23 +12,38 @@ namespace TLSharp.Core.Network
 
     public class TcpTransport : IDisposable
     {
-        private readonly TcpClient tcpClient;
-        private readonly NetworkStream stream;
+        private TcpClient tcpClient;
+        private NetworkStream stream;
         private int sendCounter = 0;
+        TcpClientConnectionHandler handler;
+        string address;
```
knocte, line 19: address is not reused, so you don't need to create a field for it

TLSharp.Core/TelegramClient.cs
```diff
@@ -56,22 +58,32 @@ public Session Session
             if (string.IsNullOrEmpty(apiHash))
                 throw new MissingApiConfigurationException("API_HASH");
 
-            if (store == null)
-                store = new FileSessionStore();
+            this.store = store ?? new FileSessionStore();
```
knocte, line 61: this change is unrelated to the task at hand

TLSharp.Core/Network/TcpTransport.cs
```diff
@@ -12,23 +12,38 @@ namespace TLSharp.Core.Network
...
 
         public TcpTransport(string address, int port, TcpClientConnectionHandler handler = null)
+        {
+            this.handler = handler;
+            this.address = address;
+            this.port = port;
+            ipAddress = IPAddress.Parse(address);
+        }
+
+        public async Task Connect()
```
knocte, line 31: if this method is async its name should have the `Async` suffix

TLSharp.Core/TelegramClient.cs
```diff
@@ -56,22 +58,32 @@ public Session Session
...
             token.ThrowIfCancellationRequested();
 
+            if (!transport.IsConnected)
+            {
+                // we must recreate the session because it might track dirty information 
+                // of a connection that maybe was disconnected, reusing that session will cause errors
+                session = Session.TryLoadOrCreateNew(store, sessionUserId);
+                await transport.Connect();
+            }
+            if (!transport.IsConnected)
```
knocte, line 84: as far as I understand this cannot happen; if the call to transport.Connect above fails, then it would throw an exception from that method

TLSharp.Core/Network/TcpTransport.cs
knocte: my bad, seems it is needed, howevr these 3 new fields can be marked as readonly

TLSharp.Core/Network/TcpTransport.cs
solarin: we can change, but also send and receive in the same class are async and are not named as such. i suggest to keep the same style.

TLSharp.Core/TelegramClient.cs
solarin: yeah should be so

TLSharp.Core/Network/TcpTransport.cs
knocte: this is a convention by Microsoft, the creators of C#; if you see any current API that doesn't fit this pattern we should rename it

solarin: ok then
guys, i think we should push this fix asap, i have been testing on a software which has been running for half a day now, perfectly handling the reconnection.

before this fix, it could last only 30 minutes.

TLSharp.Core/TelegramClient.cs
```diff
@@ -56,22 +58,30 @@ public Session Session
...
+            this.store = store ?? new FileSessionStore();
+            this.sessionUserId = sessionUserId;
 
             this.apiHash = apiHash;
             this.apiId = apiId;
             this.handler = handler;
             this.dcIpVersion = dcIpVersion;
 
-            session = Session.TryLoadOrCreateNew(store, sessionUserId);
+            session = Session.TryLoadOrCreateNew(this.store, sessionUserId);
```
knocte, line 69: can't we just remove this line above? session will be initialized again anyway in ConnectAsync

solarin: unfortunately not, because the following like needs it.

to save one initialization of session we must add extra code in the Connect method

knocte:
> guys, i think we should push this fix asap, i have been testing on a software which has been running for half a day now, perfectly handling the reconnection.
> before this fix, it could last only 30 minutes.

You make it sound like if this PR fixed connection problems. It doesn't. It just allows the developer to reconnect using the same instance of TelegramClient. You could have done the same before by creating a new TelegramClient instance.

solarin:
> guys, i think we should push this fix asap, i have been testing on a software which has been running for half a day now, perfectly handling the reconnection.
> before this fix, it could last only 30 minutes.
> 
> You make it sound like if this PR fixed connection problems. It doesn't. It just allows the developer to reconnect using the same instance of TelegramClient. You could have done the same before by creating a new TelegramClient instance.

true, but still a great improvement

knocte:
Hey Mario I wrote an alternative PR with much simpler change, wouldn't that achieve the same as this?: https://github.com/sochix/TLSharp/pull/937

solarin:
> Hey Mario I wrote an alternative PR with much simpler change, wouldn't that achieve the same as this?: #937

yes, it looks fine and easier. let me test it for a couple of hours.

davidcon:
> ! IsConnected is restored as it was before
> 
> * TcpTransport has got a new Connect method which properly initializes the underlying tcpclient, this makes it possible to reconnect in case the tcpClient gets disconnected, without restarting the application or recreating a new TelegramClient object.

Excuse me @solarin I don´t know how to use your new version. 
I have downloaded the master branch and I have compiled it. 
Then I have try to use IsConnected as it shuld be but it´s always set true. 
Do I have to use any new method to know that the client is disconected? 
Or should I have to use `await client.ConnectAsync(true)` in stead of `await client.ConnectAsync()`? 

Thank you very much for you time.

PR #937, opened 2020-04-10
Core: allow re-connecting from same TelegramClient

knocte:
If the connection dropped, there was no way to reconnect from the same
instance of TelegramClient.


solarin:
but in this way you are not changing the TcpTransport initialization in the constructor either?

solarin:
i the TcpTransport initialization, in the constructor, should be changed since we are working on this issue and we have a chance to make it better

solarin:
and this would reintroduce the issue of TelegramClient.IsConnected giving always TRUE

knocte:
> but in this way you are not changing the TcpTransport initialization in the constructor either?

why change that? we create a new transport on each ConnectAsync() call, and that's it

solarin:
> but in this way you are not changing the TcpTransport initialization in the constructor either?
> 
> why change that? we create a new transport on each ConnectAsync() call, and that's it

becauese the connection inside the constructor in TcpTransport is quite terrible 

anyways, now i am running your solution. let's see if it works on the long run

solarin:
yes this seems to be working

Issue #935, opened 2020-04-09
Connection problem with emirati number


solarin:
i can't manage to connect with my Emiratis telephone numbers.
i tried 2 numbers, both same problem, same exception.

> System.IO.EndOfStreamException: Unable to read beyond the end of the stream.
>    at System.IO.BinaryReader.FillBuffer(Int32 numBytes)
>    at System.IO.BinaryReader.ReadInt64()
>    at TLSharp.Core.Network.MtProtoPlainSender.<Receive>d__6.MoveNext() in source\repos\TLSharp\TLSharp.Core\Network\MtProtoPlainSender.cs:line 51
> --- End of stack trace from previous location where exception was thrown ---
>    at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
>    at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
>    at System.Runtime.CompilerServices.ConfiguredTaskAwaitable`1.ConfiguredTaskAwaiter.GetResult()
>    at TLSharp.Core.Auth.Authenticator.<DoAuthentication>d__0.MoveNext() in source\repos\TLSharp\TLSharp.Core\Auth\Authenticator.cs:line 39
> --- End of stack trace from previous location where exception was thrown ---
>    at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
>    at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
>    at System.Runtime.CompilerServices.ConfiguredTaskAwaitable`1.ConfiguredTaskAwaiter.GetResult()
>    at TLSharp.Core.TelegramClient.<ConnectAsync>d__13.MoveNext() in source\repos\TLSharp\TLSharp.Core\TelegramClient.cs:line 87
> --- End of stack trace from previous location where exception was thrown ---
>    at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
>    at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
>    at System.Runtime.CompilerServices.ConfiguredTaskAwaitable.ConfiguredTaskAwaiter.GetResult()
[...]

PR #938, opened 2020-04-11
address an issue with the datacenter selection...

solarin:
that could select datacenters which were not supporting standard communication

### DIFF
--- a/TLSharp.Core/TelegramClient.cs
+++ b/TLSharp.Core/TelegramClient.cs
@@ -121,6 +121,8 @@ namespace TLSharp.Core
             else
                 dcs = dcOptions.Where(d => d.Id == dcId); // any
 
+            dcs = dcs.Where(d => !d.MediaOnly);
+
             TLDcOption dc;
             if (dcIpVersion != DataCenterIPVersion.Default)
             {


solarin:
@knocte is my concern clear now why we should push the other fix first?

knocte:
not clear man, what would happen if I merge only the line that filters the MediaOnly DCs and not the session save? what happens in that case?

solarin:
it works. then when you push fix 937 it won't work. because you need the line session.save

solarin:
i don't know what else i can say, try it if you don't believe/understand/unsure of what i am saying. i have been debugging for hours and now also documenting it for many more hours. 
i might sound arrogant, but i value my time a lot.

knocte:
> but i value my time a lot.

It's not that I don't believe you, it's that I want to understand before merging.

> it works. then when you push fix 937 it won't work. because you need the line session.save

Well, then I will merge first a commit filtering the MediaOnly, whose commits message says that it fixes bug https://github.com/sochix/TLSharp/issues/935. After that, I'll modify PR 937 to include that Save() call. And we're done. And each commit fixes a separate thing. Ok with this?

TLSharp.Core/TelegramClient.cs
```diff
@@ -134,8 +136,9 @@ private async Task ReconnectToDcAsync(int dcId, CancellationToken token = defaul
             
             var dataCenter = new DataCenter (dcId, dc.IpAddress, dc.Port);
 
-            transport = new TcpTransport(dc.IpAddress, dc.Port, handler);
+            transport = new TcpTransport(dc.IpAddress, dc.Port, handler); // this will be gone after fix #937
```
solarin, line 139: if you keep this line after fix 937, it will be useless because the connectionAsync on line 140 will cause  a recreation of the TcpTransport

knocte: Right so then this has to be fixed in the PR#937, not in the same commit that does the MediaOnly filter.

TLSharp.Core/TelegramClient.cs
```diff
@@ -134,8 +136,9 @@ private async Task ReconnectToDcAsync(int dcId, CancellationToken token = defaul
...
             session.DataCenter = dataCenter;
+            session.Save(); // this is needed after fix #937
```
solarin, line 141: if you don't save the session, the new tcpTransport created inside connectAsync (firx 937), will reload the session with the old datacenter value

TLSharp.Core/TelegramClient.cs
solarin: ok then
let's go ahead.

nothing personal, ok, i am a very direct and at the same time sincere person. 

cheers

PR #972, opened 2020-08-02
TelegramClient.ResolveUsernameAsync

ArmeF97:


knocte:
Why? The consumer of TLSharp should remove the '@' character at the beginning.

ArmeF97:
The PR is not for removing the @ at the beginning, is to introduce the whole function, that was not present

knocte:
Oh sorry I thought you were introducing a new overload. Can you port your PR over to TgSharp please? TLSharp is deprecated because it uses an old layer.

ArmeF97:
I suppose you are referring to this https://github.com/nblockchain/TgSharp

Thanks. Didnt' know about this

knocte:
Yes

PR #983, opened 2020-09-29
Vote Message

DenisHidden:
Please Update Layer To able send vote for like message and channel & group vote

thanks.


knocte:
What is this? If you want to contribute to TgSharp, open a PR there, not here.

PR #987, opened 2020-10-15
Layer 109

DenisHidden:
Hello TG Team Please Update Layer To 109 For Support Send Vote 

tnx.


knocte:
If you can live with layer 108, please use TgSharp.
If you need layer 109, please make a PR in TgSharp, not here.

PR #1003, opened 2020-12-06
Added a check to make sure this value is not added twice

RandallFlagg:
Added a check to make sure this value is not added twice to the dictionary otherwise got an exception when loading. Since I don't know what is this value and why it is inside the code I added a check.


knocte:
Why is appveyor failing?

PR #1004, opened 2020-12-06
Created a .NET Core project files

RandallFlagg:
There are 2 ways to solve the issue here:
1. Comment out the duplicate values.
2. Exclude the assembly info from the project

I chose 1 just because I wanted to preserve the rest of the attributes.


knocte:
We're not accepting new features in TLSharp, only bugfixes. Please propose new features in TgSharp repo.

PR #1020, opened 2021-04-16
Client: add call for resendCode request

wildtylus:
TelegramClient: added a call to resend code request.
The call needs two parameters: 

- phoneNumber - number used within sendCode() request
- phoneCodeHash - hash received as a result from sendCode() request
Note: Telegram switches the type of code it sends during resend. If sendCode request got a Telegram message with the code, resendCode request will get SMS message with the code.
more details: https://core.telegram.org/method/auth.resendCode


maurei:
Why is this not merged?

